PR #321, opened 2014-11-01
Quadratic Minimization solver using ADMM based Proximal Algorithm

debasish83:
@dlwh @mengxr

Refactoring the QuadraticMinimizer implementation for integration into Spark mllib ALS. The solver is used to solve Quadratic minimization using several constraints (positivity, bounds, elastic net, equality and bounds). 

More details are in the following Spark JIRA and PR:

Spark JIRA https://issues.apache.org/jira/browse/SPARK-2426
Spark PR https://github.com/apache/spark/pull/2705

I also refactored NNLS to breeze since NNLS can also solve x'Hx + c'x s.t x >= 0 and it is a special case of the features that ProximalMinimizer supports.

I used jblas since that's what mllib ALS uses and so in this cleanup we also want to move ALS to breeze matrices and therefore I will clean it up.

Matlab comparisons of the proximal solver is presented over here. We are at par with MOSEK till ranks ~ 1000 but due to ADMM we have the luxury of moving x-solve to each worker later once we validate the solver.
https://github.com/debasish83/ecos/blob/master/README.md

In scala version the runtime numbers can be obtained as follows:

run-class breeze.optimize.quadratic.ProximalMinimizer 1000 1 1.0 0.99

rank=1000, equality=1.0 lambda=1.0 beta=0.99
L1regularization = lambda*beta L2regularization = lambda*(1-beta)
Generating randomized QPs with rank 1000 equalities 1
sparseQp 88.423 ms iterations 45 converged true
posQp 181.369 ms iterations 121 converged true
boundsQp 175.733 ms iterations 121 converged true
Qp Equality 2805.564 ms iterations 2230 converged true

The next steps are the following:

1. Refactor using breeze code style
2. Use breeze dense/sparse matrices
3. Move tests from main to test
4. Iterative version taking DiffFunction and not dense gram so that we can scale to large ranks and handle convex functions in mllib ALS (KL divergence loss, hinge loss, huber loss)

For 4, I will need BFGS/CG core but that's not really critical for recommendation use-cases. Basically lot of code will be re-used from breeze.optimize. Also we have a baseline direct solver for comparisons which is critical for the iterative solver validation.


debasish83:
@dlwh With quadratic problems constructed from normal gaussian as the random seed, now for unconstrained minimization bfgs results match well with lu, cg and qp.

> runMain breeze.optimize.quadratic.QuadraticMinimizer 200 1 0.1 0.99
[info] Running breeze.optimize.quadratic.QuadraticMinimizer 200 1 0.1 0.99
Generating randomized QPs with rank 200 equalities 1
Test QuadraticMinimizer, CG , BFGS and OWLQN with 200 variables and 1 equality constraints

||qp - lu|| norm 1.3392253191795917E-10 max-norm 2.6370017280896718E-11
||cg - lu|| norm 1.2825805307285098E-8 max-norm 2.668802512317825E-9
||bfgs - lu|| norm 0.20565842746229476 max-norm 0.04131523432133477
Objective lu -22.890582186874294 bfgs -22.89041818327972 qp -22.89058218685401
dim 200 lu 142.376 ms qp 61.161 ms cg 156.457 ms bfgs 1005.704 ms

Also owlqn and elastic net results from QuadraticMinimizer match well:

||owlqn - sparseqp|| norm 0.033344628875555476 inf-norm 0.006976122518636885
sparseQp 54.32 ms iters 299 owlqn 988.533 ms

In both cases QuadraticMinimizer runtime are faster.

I will get back to fixing rest of the comments. Got back last Friday. Still a bit jet-lagged !

math/src/main/scala/breeze/optimize/quadratic/NNLS.scala
```diff
@@ -0,0 +1,182 @@
...
+   *   subject to x >= 0
+   *
+   * The method used is similar to one described by Polyak (B. T. Polyak, The conjugate gradient
+   * method in extremal problems, Zh. Vychisl. Mat. Mat. Fiz. 9(4)(1969), pp. 94-112) for bound-
+   * constrained nonlinear programming.  Polyak unconditionally uses a conjugate gradient
+   * direction, however, while this method only uses a conjugate gradient direction if the last
+   * iteration did not cause a previously-inactive constraint to become active.
+   */
+  def solve(ata: DenseMatrix[Double], atb: DenseVector[Double]) : Array[Double] = {
+    wipe()
```
debasish83: I am considering adding the state in ThreadLocal...when we use this solver inside ALS, typically 100K solves are being run using the same workspace and so if the workspace is hot runtime will be better...I can do some analysis to see what happens if the workspace is created at each solve....

debasish83: @dlwh I would like to use the Iterator pattern that you use in other optimizers as that will fix the State issue but I am really confused about the pattern...say if I run BFGS with 100 iterations, am I not keeping 100 State(x, grad, val)...if x is of dimension 1M, is not the size of states will be 1M x 100 + 1M x 100 + 100...feels like that is lot of memory waste...

dlwh: It forgets the states as soon as next is called, so it's not a problem.

debasish83:
@dlwh could you please make a second pass ? NNLS is now made stateless after I compared the performance...I did not find any difference...QuadraticMinimizer still holds the memory for gram matrix since I need the updateGram API when the solver will be called from MLLIB/ML ALS code...

dlwh:
yeah, i need to lighten the FP tolerance on one of the tests.

math/src/main/scala/breeze/optimize/quadratic/NNLS.scala
```diff
@@ -0,0 +1,199 @@
...
+   * method in extremal problems, Zh. Vychisl. Mat. Mat. Fiz. 9(4)(1969), pp. 94-112) for bound-
+   * constrained nonlinear programming.  Polyak unconditionally uses a conjugate gradient
+   * direction, however, while this method only uses a conjugate gradient direction if the last
+   * iteration did not cause a previously-inactive constraint to become active.
+   */
+  def minimize(ata: DenseMatrix[Double], atb: DenseVector[Double]) : DenseVector[Double] = {
+    iterations(ata, atb).x
+  }
+
+  def iterations(ata: DenseMatrix[Double], atb: DenseVector[Double]) : State = {
```
dlwh: this doesn't actually return an iterator?

math/src/main/scala/breeze/optimize/quadratic/NNLS.scala
```diff
@@ -0,0 +1,199 @@
...
+
+    var lastNorm = 0.0
+    var iterno = 0
+    var lastWall = 0 // Last iteration when we hit a bound constraint.
+    var i = 0
+
+    while (iterno < iterMax) {
+      // find the residual
+      res := mult(ata, x)
+      axpy(-1.0, atb, res)
```
dlwh: just use -= ?

math/src/main/scala/breeze/optimize/quadratic/NNLS.scala
```diff
@@ -0,0 +1,199 @@
...
+import breeze.linalg.axpy
+import breeze.linalg.operators.OpMulMatrix
+import breeze.stats.distributions.Rand
+/**
+ * Object used to solve nonnegative least squares problems using a modified
+ * projected gradient method.
+ */
+
+class NNLS(val maxIters: Int = -1)
+          (implicit mult: OpMulMatrix.Impl2[DenseMatrix[Double], DenseVector[Double], DenseVector[Double]]) {
```
dlwh: you don't need this implicit

math/src/main/scala/breeze/optimize/quadratic/NNLS.scala
```diff
@@ -0,0 +1,199 @@
...
+        ndir = dir.dot(dir)
+      }
+
+      // terminate?
+      if (stop(step, ndir, nx)) {
+        return State(x, grad, res, iterno, true)
+      }
+      
+      // don't run through the walls
+      i = 0
```
dlwh: not important to make this change, but these days we're preferring cforRange over while loops (there are still plenty of while loops in Breeze, but in general...)

math/src/main/scala/breeze/optimize/quadratic/NNLS.scala
```diff
@@ -0,0 +1,199 @@
...
+    val iterMax = if (maxIters < 0) Math.max(400, 20 * n) else maxIters
+
+    var lastNorm = 0.0
+    var iterno = 0
+    var lastWall = 0 // Last iteration when we hit a bound constraint.
+    var i = 0
+
+    while (iterno < iterMax) {
+      // find the residual
+      res := mult(ata, x)
```
dlwh: just use * ?

math/src/main/scala/breeze/optimize/quadratic/QuadraticMinimizer.scala
```diff
@@ -0,0 +1,545 @@
...
+  def project(x: DenseVector[Double]) = {
+    var i = 0
+    while (i < x.length) {
+      if (abs(x.data(i)) <= EPS) x.data(i) = 0.0
+      i = i + 1
+    }
+    x
+  }
+
+  def iterations(q: DenseVector[Double]): State = {
```
dlwh: same, doesn't return iterator?

math/src/main/scala/breeze/optimize/quadratic/QuadraticMinimizer.scala
```diff
@@ -0,0 +1,545 @@
...
+  def setLambda(lambda: Double): QuadraticMinimizer = {
+    this.lambda = lambda
+    this
+  }
+
+  //TO DO : This can take a proximal function as input
+  //TO DO : alpha needs to be scaled based on Nesterov's acceleration
+  def setProximal(constraint: Constraint,
+                  lb: Option[DenseVector[Double]] = None,
+                  ub: Option[DenseVector[Double]] = None): QuadraticMinimizer = {
```
dlwh: i don't like setters for things like this, but if it's needed downstream, I guess it's ok.

math/src/main/scala/breeze/optimize/quadratic/Proximal.scala
```diff
@@ -0,0 +1,240 @@
...
+      x = (lIter + uIter) / 2
+      iter = iter + 1
+    }
+    x
+  }
+
+  def proxSeparable(x: DenseVector[Double], rho: Double, oracle: Double => Double, l: Double, u: Double) {
+    var i = 0
+    while (i < x.length) {
+      x.update(i, proxScalar(x(i), rho, oracle, l, u, 0))
```
dlwh: x.map(proxScalar(_, rho, oracle, l, u, 0)) ?

math/src/main/scala/breeze/optimize/quadratic/Proximal.scala
```diff
@@ -0,0 +1,240 @@
...
+}
+
+// f = huber = x^2 if |x|<=1, 2|x| - 1 otherwise
+case class ProximalHuber() extends Proximal {
+  def proxScalar(v: Double, rho: Double, oracle: Double => Double, l: Double, u: Double, x0: Double): Double = {
+    val MAX_ITER = 1000
+    val tol = 1e-8
+
+    var g: Double = 0.0
+    var x = max(l, min(x0, u));
```
dlwh: x = clip(x0, l, u)

math/src/main/scala/breeze/optimize/quadratic/NNLS.scala
debasish83: Yeah I can change it to Iterator but I did not see how the iterator pattern will improve over the while loop...Did you intend to do things like iterator.next will be run on thread1 while iterator.next.next is run on thread2 using Iterator pattern ? This is still not a asynchronous algorithm...may be we can do this change as a separate PR ?

math/src/main/scala/breeze/optimize/quadratic/NNLS.scala
debasish83: fixed it..

math/src/main/scala/breeze/optimize/quadratic/NNLS.scala
debasish83: If I take out implicit, the companion objects fails:

[error] /Users/v606014/github/breeze/math/src/main/scala/breeze/optimize/quadratic/NNLS.scala:174: missing arguments for constructor NNLS in class NNLS
[error]   def apply(iters: Int) = new NNLS(iters)
[error]                           ^
[error] /Users/v606014/github/breeze/math/src/main/scala/breeze/optimize/quadratic/NNLS.scala:185: missing arguments for constructor NNLS in class NNLS
[error] Error occurred in an application involving default arguments.
[error]     val nnls = new NNLS()

math/src/main/scala/breeze/optimize/quadratic/NNLS.scala
dlwh: The whole reason to do iterators is to allow the user to monitor the state of the optimization, the way callbacks are used in lbfgs-c (http://www.chokkan.org/software/liblbfgs/ the "progress" function in the example.) Users can then use their own convergence checks, cache intermediate results, etc. E.g. I often use the iterations method on our first order minimizers to run on a held out set to check for when to terminate the optimization.

math/src/main/scala/breeze/optimize/quadratic/NNLS.scala
dlwh: i meant you can delete that entire argument: " (implicit mult:
OpMulMatrix.Impl2[DenseMatrix[Double], DenseVector[Double],
DenseVector[Double]]) " you can delete that.

debasish83: Actually I don't need this mult at all...I cleaned all mult and use DenseMatrix[Double]*DenseVector[Double]...

math/src/main/scala/breeze/optimize/quadratic/QuadraticMinimizer.scala
debasish83: Yes...In ALS the lambda for every solve is a function of how many ratings are there for every user..therefore lambda needs to be updated for every solve

math/src/main/scala/breeze/optimize/quadratic/NNLS.scala
debasish83: I used that package but never paid attention to progress ! Yeah it is definitely useful for users...Let me update both the algorithms with the iterator pattern...

math/src/main/scala/breeze/optimize/quadratic/Proximal.scala
debasish83: I expect to do in-place modification in Proximal trait....if I do x.map, I have to return DenseVector[Double] from prox interface...

math/src/main/scala/breeze/optimize/quadratic/Proximal.scala
debasish83: clip is defined on an array but I am doing component-wise operation over here...

math/src/main/scala/breeze/optimize/quadratic/QuadraticMinimizer.scala
debasish83: I am cleaning up the setProximal API since we can add the Proximal in the constructor...setLambda API will stay for now due to ALS lambda tuning requirement per user/product...

math/src/main/scala/breeze/optimize/quadratic/Proximal.scala
dlwh: huh, ok. I'll open a bug for that, but leave as is for now, of course.

dlwh:
FWIW, I'll be unavailable from Feb 3, so I may not be able to finish merging. @gabeos could I impose on you to review this from the 2nd or 3rd on?

debasish83:
@dlwh @gabeos I updated the PR with all fixes that we discussed...Once the PR is merged it would be great if the following README can be added to breeze README. 

The QuadraticMinimizer driver randomly generate Quadratic problems using a normal gaussian...The code for quadratic problem generation is taken from PDCO...If all the columns of any arbitrary data matrix are normalized with z-score we are likely to get matrices similar to the generate quadratic problems:

runMain breeze.optimize.proximal.QuadraticMinimizer 400 1 0.1 0.99 

The first comparison is within different optimizer trying to find optimal solution of minimize 0.5x'Hx + c'x:

||qp - lu|| norm 5.537727052361776E-7 max-norm 9.626717201172141E-8
||cg - lu|| norm 1.5678987800416877E-6 max-norm 2.7266253255220363E-7
||bfgs - lu|| norm 10.883354186273674 max-norm 1.900390295935381
Objective lu -1524.8339864982945 bfgs -1524.8233992754779 qp -1524.8339860484775
dim 400 lu 151.922 ms qp 81.111 ms cg 313.938 ms bfgs 15364.397 ms

Interesting here is that breeze lu and qp (current PR) are accurate and bfgs objective is close enough. Now if I generate random gram matrix not using Rand.gaussian(0,1) but Rand.uniform then lu and qp results match but bfgs results will be way off. I valided with the f2j code from RISO project and the same behavior show up there as well.

As expect due to Cholesky and BackSolves qp (current PR) is the faster solver here. Also the dimension is low which helps Cholesky factorizations O(n^3) complexity. 

The second comparison is within Breeze OWLQN and sparseQp (current PR). SparseQp is the following elastic net formulation:
minimize 0.5x'Hx + c'x + lambda(1 - beta)*L2(x) + lambda*beta*L1(z)
s.t x = z
We use ADMM as the core algorithm but other FBS variants like FISTA is equally applicable. Due to our familiarity with ADMM we used it. For our run lambda was 0.1 and beta is 0.99

The following is the difference in solution quality within owlqn and sparseqp. They are pretty close:
||owlqn - sparseqp|| norm 0.1563477897709106 inf-norm 0.03232306408737637

Now look at the iterations (~3x lesser) and runtime (~6x faster):
sparseQp 132.196 ms iters 116 owlqn 756.228 ms iters 360

The third comparison is within posQp (current PR) and NNLS (moved from Spark optimization to breeze.optimize.linear). NNLS uses CG core and projected gradient to satisfy positivity constraint. posQp formulates the problem as follows:
minimize 0.5x'Hx + c'x + g(z) , g(z) = {z | z >= 0}
s.t x = z

The iterations of posQp and NNLS are very close to each other with similar runtimes.
posQp 66.545 ms iters 198 nnls 71.26 ms iters 226

These numbers indicate that both projected gradient and proximal algorithm are at par with each other.

Proximal algorithm machinery using ADMM core is used to solve the last two formulations. 

For boundsQp the formulation is:
minimize 0.5x'Hx + c'x + g(z) , g(z) = {z | lb <= z <= ub}
s.t x = z

Here are the runtime numbers:
boundsQp 75.906 ms iters 198 converged true
For this problem, boundsQp took 198 iterations to converge in 76 ms

For equalityQp the formulation is:
minimize  0.5x'Hx + c'x + g(z), g(z) = { z | lb <= z <= ub }
s.t Aeq x = beq

The affine constraint is added in the primal solve which allows solve general quadratic programs through this PR. For machine learning probability simplex 1'x = 1, g(z) = { z | z >= 0} is more interesting and we provide a companion object to generate the probability simplex and then QuadraticMinimizer can be used to solve for it.

Here are the runtime numbers:
Qp Equality 44.46 ms iters 52 converged true
For this problem equalityQp took 52 iters and 44 ms

gabeos:
Hey @debasish83 Thanks for the write-up. I'm super busy tomorrow, but I'll take a look within the following day or so. Will probably take me longer than David would take, but I'll do my best to be prompt. 

I think your write-up above could make a great start to documenting breeze's optimization functionality, which doesn't yet have it's own section in the GH wiki, just a small section in the Quickstart page, so after we handle the merge, I'm happy to work with you on getting that info in there, although it should probably be re-structured and edited to be easier to understand for those with less of a math/ML background.

debasish83:
@gabeos sure please take your time...It is critical to have the write-up well so that people can make use of the proximal algorithms in their projects...constraint handling in open source have been a pain so far...

This PR is focused on smaller dimensions where it is possible to explicitly construct gram matrix which can be factorized using cholesky/lu. We support dense gram matrix right now but in future I intend to support sparse gram matrices through CSCMatrix.

I am testing the nonlinear version which uses lot of Breeze goodies like BFGS, ApproximateInverseHessian and CompactHessian. It will scale to problems of ~ 10M dimension (basically bounded by LBFGS matrix size on one Node). If it works as we expect I will open up the PR. It will further show the value of breeze.optimize internals.

debasish83:
@dlwh @gabeos I needed power method to generate estimates of eigen values and I wanted to support both DenseMatrix and BFGS.ApproximateInverseHessian for power law...There was a TO DO in the code for computeRho

If you look at breeze.optimize.linear.PowerMethod, right now I am doing the following for inverse power law which is ugly. I am not sure how to get rid of this:

val ay = if (inverse)
QuadraticMinimizer.solveTriangular(A.asInstanceOf[DenseMatrix[Double]], eigenVector.asInstanceOf[DenseVector[Double]]).asInstanceOf[T]
else mult(A, eigenVector)

Basically adding a require statement that inverse power law is only supported for DenseMatrix[Double] and DenseVector[Double] is fine...If there are other ways to clean it please suggest... 

solveTriangular is basically a back solve on cholesky factor...usage is in PowerMethodTest.scala

val pmInv = new PowerMethod[DenseVector[Double], DenseMatrix[Double]](10, 1e-5, true)
val R = cholesky(gram).t
val eigenApprox = 1.0/pmInv.eigen(init, R)

For BFGS.ApproximateInverseHessian, running the normal power method will give me minEigen

I construct PQN.CompactHessian from BFGS history and then running the normal power method will given me maxEigen. So we are covered on those.

debasish83:
@dlwh discussed with @mengxr and @fommil. Merging this PR will bring NNLS to breeze and that will help us removing all jblas dependencies from mllib/ml ALS and use breeze + mllib data structures. Right now ALS is the only code which uses jblas due to NNLS.

math/src/main/scala/breeze/optimize/linear/NNLS.scala
```diff
@@ -0,0 +1,203 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
```
dlwh: fix license headers; this is not an apache project. Just use your name like I do in the files I've made.

debasish83: fixed all the license headers and added tests for the minimize API that uses updateGram.

math/src/main/scala/breeze/optimize/linear/PowerMethod.scala
```diff
@@ -0,0 +1,64 @@
...
+    y *= 1.0/norm1
+    if (lambda < 0.0) y *= -1.0
+    State(lambda, y, 0, false)
+  }
+
+  def iterations(y: T,
+                 A: M): Iterator[State] = Iterator.iterate(initialState(y, A)) { state =>
+    import state._
+    val ay = if (inverse)
+      QuadraticMinimizer.solveTriangular(A.asInstanceOf[DenseMatrix[Double]], eigenVector.asInstanceOf[DenseVector[Double]]).asInstanceOf[T]
```
debasish83: @dlwh For DenseMatrix[Double] I am using inverse power method here...I can catch ClassCastException and add a WARN that inverse power method right now is only defined for DenseMatrix[Double]...Is there an elegant way to fix it ? Down the line I would like to support SparseMatrix[T] and DenseMatrix[T] but right now LU and cholesky are both using Double and so I could not specialize QuadraticMinimizer on Float, Double

dlwh: I mean, there's just no reason to make PowerMethod generic in M and T if it actually isn't generic.

debasish83: I kept PowerMethod generic in M and T (since I want to use it over LBFGS.ApproximateInverseHessian and PQN.CompactHessian if required). For InversePowerMethod I introduced a new class specialized on DenseMatrix[Double], DenseVector[Double] to be used from QuadraticMinimizer...Now it is clean.

math/src/main/scala/breeze/optimize/linear/InversePowerMethod.scala
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Inverse Power Method specialized on DenseVector[Double], DenseMatrix[Double]
+ * For usage see breeze.optimize.proximal.QuadraticMinimizer rho computation
+ * @author debasish83
+ */
+class InversePowerMethod(maxIterations: Int = 10, tolerance: Double = 1E-5)
+                        (implicit space: MutableInnerProductModule[DenseVector[Double], Double],
+                         mult: OpMulMatrix.Impl2[DenseMatrix[Double], DenseVector[Double], DenseVector[Double]])
+  extends PowerMethod[DenseVector[Double], DenseMatrix[Double]] {
```
dlwh: just noticed this extends PowerMethod. I generally don't like extending concrete classes.

dlwh: if it's specialized you don't need the space and mult implicit arguments.

math/src/main/scala/breeze/optimize/proximal/QuadraticMinimizer.scala
```diff
@@ -0,0 +1,521 @@
...
+  //u is the langrange multiplier
+  //z is for the proximal operator application
+
+  def initialState(nGram: Int) = {
+    var R: DenseMatrix[Double] = null
+    var pivot: Array[Int] = null
+
+    //Dense cholesky factorization if the gram matrix is well defined
+    if (linearEquality > 0) {
+      val lu= LU(wsH.asInstanceOf[DenseMatrix[Double]])
```
dlwh: why is this cast here?

math/src/main/scala/breeze/optimize/linear/PowerMethod.scala
dlwh: Sounds good. Made a few more comments.

math/src/main/scala/breeze/optimize/linear/InversePowerMethod.scala
debasish83: fixed space and implicit but do u mean add an object PowerMethod and then define inverse similar to what you do for DiffFunction.withL2Regularization for example ?

math/src/main/scala/breeze/optimize/proximal/QuadraticMinimizer.scala
debasish83: cleaned

math/src/main/scala/breeze/optimize/linear/InversePowerMethod.scala
debasish83: I added PowerMethod.inverse and cleaned up InversePowerMethod class

math/src/main/scala/breeze/optimize/linear/NNLS.scala
```diff
@@ -0,0 +1,189 @@
...
+
+class NNLS(val maxIters: Int = -1) {
+  type BDM = DenseMatrix[Double]
+  type BDV = DenseVector[Double]
+
+  case class State(x: BDV, grad: BDV, dir: BDV, lastDir: BDV, res: BDV,
+                   lastNorm: Double, lastWall: Int, iter: Int, converged: Boolean)
+
+  // find the optimal unconstrained step
+  def steplen(ata: BDM, dir: BDV, res: BDV,
```
dlwh: any reason to not make all these helper methods private?

math/src/main/scala/breeze/optimize/linear/NNLS.scala
```diff
@@ -0,0 +1,189 @@
...
+    ((step.isNaN) // NaN
+      || (step < 1e-7) // too small or negative
+      || (step > 1e40) // too small; almost certainly numerical problems
+      || (ndir < 1e-12 * nx) // gradient relatively too small
+      || (ndir < 1e-32) // gradient absolutely too small; numerical issues may lurk
+      )
+  }
+
+  /**
+   * Solve a least squares problem, possibly with nonnegativity constraints, by a modified
```
dlwh: can you move this to iterations or minimize or something?

math/src/main/scala/breeze/optimize/linear/NNLS.scala
```diff
@@ -0,0 +1,189 @@
...
+
+      // find the residual
+      res := ata * x
+      res -= atb
+      grad := res
+
+      // project the gradient
+      var i = 0
+      while (i < n) {
+        if (grad.data(i) > 0.0 && x.data(i) == 0.0) {
```
dlwh: in general we prefer grad(i) or grad.unsafeValueAt(i), unless this is making a noticeable difference. Also, cforRange is better than a while loop.

math/src/main/scala/breeze/optimize/linear/NNLS.scala
```diff
@@ -0,0 +1,189 @@
...
+    var nnlsTime = 0L
+    while(i < numSolves) {
+      val ata = QpGenerator.getGram(problemSize)
+      val atb = DenseVector.rand[Double](problemSize, Rand.gaussian(0,1))
+      val startTime = System.nanoTime()
+      nnls.minimize(ata, atb)
+      nnlsTime = nnlsTime + (System.nanoTime() - startTime)
+      i = i + 1
+    }
+    println(s"NNLS problemSize $problemSize solves $numSolves ${nnlsTime/1e6} ms")
```
dlwh: in general, I prefer to use the scala-logging libraries

dlwh: oh, nvm, this is a main method.

math/src/main/scala/breeze/optimize/linear/PowerMethod.scala
```diff
@@ -0,0 +1,83 @@
...
+class PowerMethod[T, M](maxIters: Int = 10,tolerance: Double = 1E-5)
+                       (implicit space: MutableInnerProductModule[T, Double],
+                        mult: OpMulMatrix.Impl2[M, T, T]) extends SerializableLogging {
+
+  import space._
+
+  case class State(eigenValue: Double, eigenVector: T, iter: Int, converged: Boolean)
+
+  //memory allocation for the eigen vector result
+  def normalize(y: T) : T = {
```
dlwh: It you require that space is a MutableInnerProductSpace instead, you can use the built in normalize instead.

math/src/main/scala/breeze/optimize/linear/PowerMethod.scala
```diff
@@ -0,0 +1,83 @@
...
+    iterations(y, A).last
+  }
+
+  def eigen(y: T, A: M): Double = {
+    iterateAndReturnState(y, A).eigenValue
+  }
+}
+
+object PowerMethod {
+  def inverse(maxIters: Int = 10, tolerance: Double = 1E-5) = new PowerMethod[DenseVector[Double], DenseMatrix[Double]](maxIters, tolerance) {
```
dlwh: always annotate return type when anonymous derived classes are returned; otherwise Scala infers the structural refinement, which slows things down.

math/src/main/scala/breeze/optimize/linear/NNLS.scala
debasish83: added private to both NNLS and QuadraticMinimizer

math/src/main/scala/breeze/optimize/linear/NNLS.scala
debasish83: this has been called from 2 places in iterations and that's why I defined it as a function

math/src/main/scala/breeze/optimize/linear/NNLS.scala
debasish83: added cforRange

math/src/main/scala/breeze/optimize/linear/PowerMethod.scala
debasish83: LBFGS uses implicit space: MutableInnerProductModule[T, Double] and that's why I used it...can we change LBFGS to MutableInnerProductSpace ?

math/src/main/scala/breeze/optimize/linear/PowerMethod.scala
debasish83: fixed

math/src/main/scala/breeze/optimize/linear/NNLS.scala
dlwh: i meant the comments

math/src/main/scala/breeze/optimize/linear/PowerMethod.scala
dlwh: `MutableInnerProductModule[T, Double] :> MutableInnerProductSpace[T, Double]`, but it should be fine to change LBFGS to use MutableInnerProductSpace, since we always have a Field for Double.

math/src/main/scala/breeze/optimize/proximal/QuadraticMinimizer.scala
```diff
@@ -0,0 +1,521 @@
...
+     * Aeq is l x rank
+     * H is rank x rank
+     * wsH is a quasi-definite matrix
+     * [ P + rho*I, A' ]
+	   * [ A        , 0  ]
+	  */
+    val ws = DenseMatrix.zeros[Double](n, n)
+    val transAeq = Aeq.t
+
+    for (row <- 0 until Aeq.rows)
```
dlwh: couldn't this be ws(nGram until (nGram + Aeq.rows), 0 until Aeq.cols) := Aeq ?

math/src/main/scala/breeze/optimize/proximal/QuadraticMinimizer.scala
```diff
@@ -0,0 +1,521 @@
...
+
+  def iterations(q: DenseVector[Double],
+                 rho: Double): Iterator[State] = Iterator.iterate(initialState(nGram)) { state =>
+    import state._
+
+    //scale will hold q + linearEqualities
+    val scale = DenseVector.zeros[Double](n)
+    val convergenceScale = sqrt(n)
+
+    //scale = rho*(z - u) - q
```
dlwh: why not use the vectorized notation here?

math/src/main/scala/breeze/optimize/proximal/QuadraticMinimizer.scala
```diff
@@ -0,0 +1,521 @@
...
+ *
+ * @param nGram rank of dense gram matrix
+ * @param proximal proximal operator to be used
+ * @param Aeq rhs matrix for equality constraints
+ * @param beq lhs constants for equality constraints
+ * @author debasish83
+ */
+class QuadraticMinimizer(nGram: Int,
+                         proximal: Proximal = null,
+                         Aeq: DenseMatrix[Double] = null,
```
dlwh: maybe we already talked about this, but why are Aeq and beq constructor parameters and not part of the call to minimize?

math/src/main/scala/breeze/optimize/proximal/QuadraticMinimizer.scala
```diff
@@ -0,0 +1,521 @@
...
+
+  if(linearEquality > 0)
+    require(beq.length == linearEquality, s"QuadraticMinimizer linear equalities should match beq vector")
+
+  val n = nGram + linearEquality
+
+  //TO DO: Tune alpha based on Nesterov's acceleration
+  val alpha: Double = 1.0
+
+  val wsH = if (linearEquality > 0) {
```
dlwh: why not make this a local and thread it through the calls to iterations etc?

math/src/main/scala/breeze/optimize/proximal/QuadraticMinimizer.scala
debasish83: I allocate the wsH during construction and I don't want to change it...the size of wsH is a function of Aeq...Building wsH is the most time consuming in the whole flow and that's why I support updateGram so that users can keep reusing the space...for mllib ALS it was a significant runtime advantage since there are 100K calls with the same rank where updateGram could be called to update wsH..for all the solves Aeq and beq are fixed...

math/src/main/scala/breeze/optimize/proximal/QuadraticMinimizer.scala
debasish83: you mean make wsH threadLocal ?

math/src/main/scala/breeze/optimize/proximal/QuadraticMinimizer.scala
dlwh: i guess I don't understand why def minimize() doesn't take the constraints as an argument, basically. Are you calling this repeatedly with different rho's and q's?

debasish83: I am making a object with Aeq and beq and then doing 100K solves by updating gram matrix through the API, QuadraticMinimizer.updateGram(row, col, value) and then calling minimize with a different rho and q...If I reconstruct a DenseMatrix each time minimize is called it will affect runtime..

debasish83: Aeq and beq in this case is probability simplex, 1'x = 1 and x >=0 I satisfy through proximal...

dlwh: gotcha. ok, I'll do one more pass and I think we're about done

math/src/main/scala/breeze/optimize/proximal/QuadraticMinimizer.scala
dlwh: could you maybe add a note about what wsH is meant to represent?

math/src/main/scala/breeze/optimize/linear/PowerMethod.scala
debasish83: I added MutableInnerProductModule to PowerMethod as well as BFGS...normalize(y) does not work in PowerMethod...gives following error:

[error] /Users/v606014/github/breeze-qp/math/src/main/scala/breeze/optimize/linear/PowerMethod.scala:41: could not find implicit value for parameter impl: breeze.linalg.normalize.Impl[T,VR]
[error]     val ynorm = normalize(y)

dlwh: Make it a MutableInnerProductVectorSpace. Modules are backed by Semirings,
so they can't do division

math/src/main/scala/breeze/optimize/proximal/QuadraticMinimizer.scala
debasish83: I updated it with the syntax...

math/src/main/scala/breeze/optimize/proximal/QuadraticMinimizer.scala
debasish83: I added the note...due to supporting the quasi definite system we can solver things as follows:
0.5x'Hx + c'x + g(z)
s.t Aeqx = beq 
g(z) here can be any proximal/projection operator...lb <= x <= ub for example brings this solver close to Matlab QuadProg

math/src/main/scala/breeze/optimize/proximal/QuadraticMinimizer.scala
debasish83: Could you please confirm that I am not allocating any memory other than "scale" in the following vectorized notation code ?
    val scale = DenseVector.zeros[Double](n)
    scale(0 until z.length) := (z-u)*rho
    scale(0 until z.length) += q*(-1.0)
    if(linearEquality > 0) scale(nGram until nGram + beq.length) := beq

debasish83: If (z - u)rho and q(-1.0) is allocating new DenseVector I will rather stick with the non-vectorized notation...

math/src/main/scala/breeze/optimize/proximal/QuadraticMinimizer.scala
debasish83: I don't want to make wsH threadLocal since this code is supposed to run on a process with multiple threads so that we can make use of multithreaded BLAS / MKL to speed up all the matrix vector computations. Next version will move from dense gram to sparse gram (basically gradually move to a interior point method over time if required)

math/src/main/scala/breeze/optimize/proximal/QuadraticMinimizer.scala
dlwh: z - u allocates, as does x * rho

In general I suggest axpy(a, x, y), which is y += (x * a), but it's carefully optimized and allocation free.

In any event, you want to move to cfor if you're concerned about allocations. Right now you're doing 1 Integer boxed allocation for all of z.length

debasish83: I will move to cfor...

math/src/main/scala/breeze/optimize/proximal/QuadraticMinimizer.scala
dlwh: I didn't mean thread local. I meant, why is this a public instance variable? It should be private, at the minimum. 

What I originally meant was make it local to the iterations method, and pass ("thread") it through  the various calls. But I don't think that's a good idea anymore.

math/src/main/scala/breeze/optimize/proximal/QuadraticMinimizer.scala
dlwh: (that's not quite true since a lot of Integers are cached, but still

math/src/main/scala/breeze/optimize/proximal/QuadraticMinimizer.scala
debasish83: Is it possible to use a async somehow on updateGram because that's the blocking operation right...multiple threads can update the gram using updateGram...

math/src/main/scala/breeze/optimize/linear/PowerMethod.scala
debasish83: I tried implicit space: MutableInnerProductVectorSpace[T, Double] but even that fails...same error shows up...

math/src/main/scala/breeze/optimize/proximal/QuadraticMinimizer.scala
debasish83: Agreed...I made it private...for async I have to think more on it...Idea here is that each process runs multithread QuadraticMinimzer using best threaded BLAS and we use a consensus algorithm from Spark that's operated by the Master..

debasish83:
I took care of all the comments...could you please take another pass ? I could not clean up normalize from PowerMethod by using implicit space: MutableInnerProductModule[T, Double] or implicit space: MutableInnerProductVectorSpace[T, Double]...any pointers would be great...the error is still:

[error] /Users/v606014/github/breeze-qp/math/src/main/scala/breeze/optimize/linear/PowerMethod.scala:41: could not find implicit value for parameter impl: breeze.linalg.normalize.Impl[T,VR]
[error] val ynorm = normalize(y)

debasish83:
I will open up a bug for normalize can we please merge this ? NonlinearMinimizer will need more large scale experiments before I am confident about it's scalability (basically 1 - 2 weeks more)

dlwh:
Thanks for putting so much work into this and contributing it!

debasish83:
Are the snapshot artifacts available on maven ? I wanted to clean up NNLS from mllib optimization and jblas dependencies...also update the ml.ALS code base with user and product constraints...

PR #340, opened 2015-01-03
stopping conditions and iteration count of truncated newton

leepei:
1. Following the original implementation of tron to add stopping conditions of truncated newton when the function value, the actual reduction or the predicted reduction is too small to prevent numerical errors.

2. When the step is rejected, do not increase the iteration count for when people specifying a max_iter, they are expecting the number of successful steps.


dlwh:
Could you fix the compile error? Thanks!

leepei:
The new commit should be compilable, and the unsuccessful test seemed to be unrelated to what I modified...

dlwh:
yeah there's some randomness in the tests i haven't totally stomped out yet. Thanks!

PR #347, opened 2015-01-13
Make gradient tester print a real percentage

jacobandreas:
If all dimensions tested are correct, GradientTester currently reports "1.000% ok". Under English number format conventions this should be "100.0%".


dlwh:
lol.

PR #349, opened 2015-01-16
FirstOrderMinimizer#infiniteIterations for custom stopping criteria

nightscape:
Hi David,

we have a PQN use case where the initial solution can be quite bad but the gradient is not very steep.
The FirstOrderMinimizer terminates without even doing a single iteration due to [the given convergence checks](https://github.com/crealytics/breeze/blob/master/math/src/main/scala/breeze/optimize/FirstOrderMinimizer.scala#L62).
I would like to be able to let the optimization run for at least n iterations before even checking the convergence criteria.

So this PR exposes an infiniteIterations method that can be used for custom stopping criteria.

Best
  Martin

### DIFF
--- a/math/src/main/scala/breeze/optimize/FirstOrderMinimizer.scala
+++ b/math/src/main/scala/breeze/optimize/FirstOrderMinimizer.scala
@@ -108,10 +108,10 @@ abstract class FirstOrderMinimizer[T, DF<:StochasticDiffFunction[T]](maxIter: In
      f.calculate(x)
   }
 
-  def iterations(f: DF,init: T): Iterator[State] = {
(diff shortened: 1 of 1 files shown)


nightscape:
There is one unrelated test failure which is due to some problem with DenseVector transposing.

dlwh:
Hey Martin, 

Thanks!

I'm getting a lot of complaints about convergence criteria these days. Could I impose on you to maybe experiment with tweaking them to something that works for you?

@eidonfiloi, @debasish83 same for you. Happy to take a PR at this point.

nightscape:
Hi David,

I'm not at all an expert in this field, but once I find something that works for me, I'll try to generalize it :)
Right now just forcing the optimizer to do at least n steps works reasonably well.

PR #353, opened 2015-01-27
Optimization Space Unboxed, Reader Fixes, and CSCMatrix constuctor

gabeos:
Minor changes (with a super-messy commit history... sorry..) 

The major benefits of the PR are:

* faster optimization spaces
* Bugfixes in CSVReader, TextReader

Please review, partially this is just to bring my master in line with upstream for future commits, but if its preferable, I can try to figure out how to squash all the commits to the important ones, since a lot of the changes are things I did and then reverted for this..


dlwh:
thanks!

math/src/main/scala/breeze/linalg/CSCMatrix.scala
```diff
@@ -45,9 +45,6 @@ class CSCMatrix[@spec(Double, Int, Float, Long) V: Zero] private[linalg] (privat
   extends Matrix[V] with MatrixLike[V, CSCMatrix[V]] with Serializable {
 
 
-  def this(data: Array[V],rows: Int, cols: Int, colPtrs: Array[Int],rowIndices: Array[Int]) =
```
mengxr: If we delete this constructor, there is no public constructor left for `CSCMatrix`.

Issue #324, opened 2014-11-12
thin QR factorisation


darrenjw:
The thin QR of an n x p matrix consists of the first p columns of Q and the first p rows of R, relative to the "usual" factorisation. In many statistical applications n >> p, so it is inefficient to first construct the full QR, which is rarely required.
Should be possible to construct with minor modification of existing function, using calls to lapack.dgeqrf, etc.

PR #361, opened 2015-02-03
Reduced mode for QR factorization

nikdon:
Reduced mode for QR factorization of matrix A(M, N) that allows to return Q and R with dimensions (M, K) and (K, N) with K = min(M, N) correspondently. Refer to issue #324


dlwh:
Thanks so much!

PR #362, opened 2015-02-04
Create empty DenseVector with given length

rakeshchalasani:
Creates an empty DenseVector with given length eg: 
`val v = new DenseVector[Double](10)`

There is a similar constructor for DenseMatrix already, added similar functionality to DenseVector now.


dlwh:
thanks!

Issue #303, opened 2014-08-28
Issues with SVD (constructing full matrices and lack of float implementation)


eiennohito:
Is there a reason why svd uses LAPACK's "A" method instead of "S"? Using former would decrease amount of used memory in case of large matrices.
Refer:
http://www.netlib.no/netlib/lapack/double/dgesdd.f

And there is no implementation of function for floats. I can do it if there will be answer for my first question.

PR #363, opened 2015-02-10
SVD that returns left and right vectors with reduced sizes

nikdon:
Reduced mode for SVD of real matrix A(M, N) that allows to return U and V.t with dimensions (M, K) and (K, N) with K = min(M, N) correspondently.

Example of use:

```scala
val m = DenseMatrix((2.0,4.0),(1.0,3.0),(0.0,0.0),(0.0,0.0))
val SVD(u, s, vt) = svd.reduced(m)
```

Refer to issue #303


dlwh:
looks great, thanks!

PR #364, opened 2015-02-11
NonlinearMinimizer using Projection and Proximal Operators

debasish83:
@dlwh @mengxr @dbtsai 

This PR introduces ADMM based Proximal algorithm to solve problems of the following structure
minimize f(x) + g(z)
s.t x = z
FISTA is also a candidate to solve this problem but we used ADMM since FISTA can't handle affine constraints natively. 

Currently f(x) can be convex/nonlinear function. So far my stress tests are using convex functions (least square regression and logistic regression) but down the line we will come to nonlinear loss (DSGD formulation from matrix factorization or sigmoid function from neural net). We call it centralized ADMM since ADMM loop is still maintained on the Master node. Once centralized ADMM works, we will generalize it to Proximal Server (Parameter server with Proximal Algorithms).

Parameter Server with Proximal Algorithms is motivated by Mu's comparison of Blocked Proximal Gradient with Google's Distributed BFGS code:
http://www.cs.cmu.edu/~muli/file/parameter_server_osdi14.pdf

We have two variants of ADMM iterations right now (experimentations are ongoing). One hot-starts on x from previous ADMM iterations while the other hot-starts on both x and BFGS history from previous ADMM iteration. New LBFGS API is added that returns state to facilitate BFGS history preservation.

runMain breeze.optimize.proximal.NonlinearMinimizer 200 0.4 0.99

We generate random data matrix (A) of size 200x200 from normal gaussian (0 mean 1 stddev). Label vector is generated from normal gaussian as well. If label > 0.5 we assign it to 1.0 else 0.0

Lambda = 0.4 ElasticNet beta = 0.99
lambdaL1 = 0.4*0.99
lambdaL2 = 0.4*0.01

Linear Regression:

Comparison 1: NonlinearMinimizer (nl) vs QuadraticMinimizer (qp)

NonlinearMinimizer takes f(x) = ||Ax - b||_2^{2} df/dx = 2(Ax - b)*A 
QuadraticMinimizer takes gram=A'A, linear term= A'b

Solution quality:
||qp - nl|| norm 0.16799969531758605 max-norm 0.034283192270685436
||bfgs - nl|| norm 0.25715611616302037 max-norm 0.04888455483124021

Objective:

Objective qp -55.00000000000021 nl -54.999725074409454 bfgs -54.99983413525827

nl and bfgs are both using BFGS for the primal solve

Comparison 2: NonlinearMinimizer with g(z)=ProximalL1 (nlSparse/nlSparseHistory) vs OWLQN vs QuadraticMinimizer with g(z)=ProximalL1 (sparseQp)

Runtime:
owlqn 165.178 ms iters 143 sparseQp 32.87 ms iters 59

Solution Quality:
||sparseQp - owlqn|| norm 0.14861916579656498 inf-norm 0.03034052809134663
owlqnObj -47.42068811893944 sparseQpObj -47.45927392747443

nlSparse 1479.666 ms iters 195 nlSparseHistory 407.347 ms iters 429
||owlqn - nlSparse|| norm 0.1460259470192177 inf-norm 0.029597210620735503
||owlqn - nlSparseHistory|| norm 0.14649731508771383 inf-norm 0.030278923124194383
nlSparseObj -47.457604531534805 nlSparseHistoryObj -47.45480418795271

NonlinearMinimizer is ~2X slower than OWLQN but if you look at objective, it converges to better solution.

Comparison 3: NonlinearMinimizer with g(z) = ProximalBounds(0 <= z <= 1.0) (nlBox) vs PQN vs QuadraticMinimizer with g(z) = ProximalBounds (qpBox)

qpBox 6.769 ms iters 57
nlBox 10182.424 ms iters 2282 nlBoxHistory 1269.134 ms 2271
pqnBox 272.064 ms iters 28

||qpBox - nlBox|| norm 0.0022824265880237215 inf-norm 4.694687018757908E-4
||qpBox - nlBoxHistory|| norm 0.0023190292089652767 inf-norm 4.773077547529439E-4
||qpBox - nlBox|| norm 0.0021390982826103347 inf-norm 5.634071605518931E-4

qpBoxObj -31.130008148965725 nlBoxObj -31.209892573229023 nlBoxHistoryObj -31.20915524772166 pqnBoxObj -31.127492181959255

Here also we are 5X slower than PQN but our solution is better compared to PQN.
 
Logistic Regression:

Comparison 1: ElasticNet formulation with OWLQN and NonlinearMinimizer with g(z)=ProximalL1 (nlLogistic)

Runtime:

owlqn 59.57 ms iters 62 nlLogistic 185.436 ms iters 33

Solution difference:

||owlqn - nlLogistic|| norm 0.41682550361219334 inf-norm 0.10509790307392614

owlqnLogistic objective 30.83266520992968
nlLogistic objective 30.817377544996752

We are ~3X compared to OWLQN but again the objective is better. So NonlinearMinimizer is producing better solution. Also see that the solutions are completely different. I have looked at the solutions. Both are sparse.

Comparison 2: Logistic Regression + L2 regularization with Bounds using PQN and NonlinearMinimizer with g(z) = ProximalBox (nlBox) 

pqnBox 226.606 ms iters 29 nlBox 1283.92 ms iters 263
pqn loss 57.56457339692869
nl loss 57.50013311157119

We are ~ 5X slower than PQN but producing better solution quality.

Next steps:

1. Runtime optimization to bring centralized ADMM closer to projected gradient algorithms (OWLQN, PQN)
2. Add proximal algorithm for probability simplex 1'x = 1, x >=0 (Centralized ADMM can solve it but not OWLQN/PQN)
3. Formulate MLE loss in mllib ALS and solve it using NonlinearMinimizer for PLSA formulation
4. Code refactoring and use Iterator pattern.
5. Test-cases showing up the objective improvement compared to OWLQN/PQN.

I wanted to open it up for early reviews. QuadraticMinimizer as mentioned here is the following PR:
https://github.com/scalanlp/breeze/pull/321


math/src/main/scala/breeze/optimize/LBFGS.scala
```diff
@@ -81,6 +81,39 @@ class LBFGS[T](maxIter: Int = -1, m: Int=10, tolerance: Double=1E-9)
...
+  /**
+   * Get the maximum approximate eigen value of the quadratic approximation convex function specified
+   * through DiffFunction by doing a inverse power iteration using the CompactHessian representation
+   * generated from ApproximateInverseHessian, get the largest eigenvalue e
+   *
+   * @param init initial guess for the eigen vector
+   * @param state the current state of the optimization
+   * @return maximum eigen value
+   */
+  def maxEigen(init:T, state: State) : Double = {
```
debasish83: @dlwh minEigen works fine but in maxEigen I am hacking it. From the BFGS history, I am trying to build the CompactHessian to run power iterations but it fails big time....Any pointers will be really helpful...

debasish83:
I fixed the failure in maxEigen but I still think it has issues. My hope was that with 10-20 iterations of BFGS we will build a good surface for the function f(x) and then minEigen and maxEigen will give us the bounds for rho as we start the proximal updates. I am using rho = 1.0 right now but in QuadraticMinimizer rho tuning with eigen values helped improving convergence. 

To test this right now I solve ||Ax - b||_2^{2} using BFGS and when converged I run LBFGS.minEigen and LBFGS.maxEigen and compare them with min(eigSym(A'A).eigenValues) and max(eigSym(A'A).eigenValues).

It's not working as I expected. 

minEigen 9.606173268621757E-6 LBFGS.minEigen -1.0165119664756253 
maxEigen 765.2710923978012 LBFGS.maxEigen 1.0000000000000002

For maxEigen power method is stopping after 1 iteration so there are still bugs in the way I construct CompactHessian from LBFGS.ApproximateInverseHistory.

dlwh:
So, first, this is very cool! I think this could be a great addition to Breeze!

A few high level comments. I'll poke around in the source as I get a chance.

1) Have you tried turning down the tolerance in LBFGS/OWLQN/PQN? I wonder if the differences in quality are from early stopping?  If not, would it maybe make sense to warm start from one of these (probably PQN as it's more general?)
2) Why can't PQN do simplex projection? E.g. http://arxiv.org/abs/1309.1541 should give the necessary projection operator?
3) Low priority, but in general I'm trying to move benchmark methods to the benchmark subproject.

NOTICE
```diff
@@ -1 +1,17 @@
...
+===============================================================================
+
+Proximal algorithms outlined in Proximal.scala (package breeze.optimize.quadratic)
+are based on https://github.com/cvxgrp/proximal (see LICENSE for details) and distributed with 
+Copyright (c) 2014 by Debasish Das (Verizon), all rights reserved.
+
+===============================================================================
+
+QuadraticMinimizer class in package breeze.optimize.proximal is distributed with Copyright (c)
+2014, Debasish Das (Verizon), all rights reserved.
```
dlwh: FWIW, all rights reserved has no legal meaning in modern copyright law.

math/src/main/scala/breeze/optimize/LBFGS.scala
```diff
@@ -81,6 +81,41 @@ class LBFGS[T](maxIter: Int = -1, m: Int=10, tolerance: Double=1E-9)
...
+  /**
+   * Get the minimum approximate eigen value of the quadratic approximation of the convex function
+   * specified through DiffFunction through power iteration on ApproximateInverseHessian,
+   * get the largest eigenvalue e and return 1/e
+   *
+   * @param init initial guess for eigen vector
+   * @param state the current state of the optimization
+   * @return minimium eigen eigen value
+   */
+  def minEigen(state: State, init: T) : Double = {
```
dlwh: wouldn't it make more sense for these to be methods on LBFGS' companion or something?

math/src/main/scala/breeze/optimize/ProjectedQuasiNewton.scala
```diff
@@ -184,5 +185,14 @@ object ProjectedQuasiNewton {
       (f, g)
     }
   }
+
+  implicit def multiplyCompactHessian[T](implicit vspace: MutableInnerProductModule[T, Double]): OpMulMatrix.Impl2[CompactHessian, T, T] = {
```
dlwh: this definitely has to go.

math/src/main/scala/breeze/optimize/LBFGS.scala
dlwh: actually, that solves the issue with the DenseVector casting. If it's on the LBFGS companion, you can just force it to be DenseVector[Double]

math/src/main/scala/breeze/optimize/linear/PowerMethod.scala
```diff
@@ -0,0 +1,78 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
```
dlwh: also, definitely not licensed to the ASF. Just put your name (c) 2015 like how I (usually) do it.

math/src/main/scala/breeze/optimize/proximal/NonlinearMinimizer.scala
```diff
@@ -0,0 +1,536 @@
...
+ * g(x) represents the following constraints
+ *
+ * 1. x >= 0
+ * 2. lb <= x <= ub
+ * 3. L1(x)
+ * 4. Aeq*x = beq
+ * 5. aeq'x = beq
+ * 6. 1'x = 1, x >= 0 ProbabilitySimplex from the reference Proximal Algorithms by Boyd et al.
+ *
+ * f(x) can be either a convex or a non-linear function.
```
dlwh: I don't know what you mean by "convex" or "non-linear function", doesn't the union of those two sets include all functions?

dlwh:
re: maxEigen, I really don't know enough about what's going on to say what's wrong, other than to speculate that the approximation may not be good enough for your purposes? (I didn't write the PQN code.) Have you tried using an EmpiricalHessian?

math/src/main/scala/breeze/optimize/proximal/NonlinearMinimizer.scala
debasish83: For convex / quadratic function bfgs + line search is good...my guess is for nonlinear like sigmoid we will need a cg based solution....may be use tron with empirical hessian...have not tested any non-convex yet...let me say that f(x) is convex

NOTICE
debasish83: PQN can do simplex projection but it does not separate f(x) and g(z)....what we really want to do is to solve f (x) through a blocked cyclic coordinate descent using bfgs and satisfy g (z) through proximal operator...that's our version of parameter server (distributed solver)...I am still thinking if we can replace coordinate descent with distributed bfgs by using some tricks...if I use owlqn or pqn I have to change code in all of them...

debasish83:
Is my CompactHessian construction from LBFGS history correct ? I thought there are bugs in it...

NOTICE
debasish83: Also I looked at the paper...that's the same algorithm I am implementing....it is a proximal algorithm....may be we can plugin to pqn as well....i dont know pqn that well...

debasish83: I read the PQN paper and it is possible to stick all the proximal operators inside PQN...I have tested bounds with PQN already and it works well compared to default ADMM. 

If it holds for ProximalL1() compared to OWLQN then I am good to use PQN as the core of NonlinearMinimizer...PQN right now accepts closure of form DenseVector[Double] => DenseVector[Double]. Would it be fine if I change the signature to (x: DenseVector[Double], rho: Double) => DenseVector[Double] ? 

This is in-tune to generic proximal algorithms:
minimize g(x) + rho||x - v||_{2}^{2}

g(x) can be constraints here: x \in C
g(x) can be L1 here as well through soft-thresholding I think

dlwh: I'm totally open to that, but I'm not quite sure how it will be used. PQN
won't manipulate it, right? Why not have your driver specify it in a new
instance of PQN?

-- David

debasish83: Seems PQN need projection operators and not really proximal....Right now L1 is the only constraint that's not working but I believe implementing L1 and probability simplex from this reference will solve it http://www.magicbroom.info/Papers/DuchiShSiCh08.pdf

box and bounds are working fine

debasish83: @mengxr 
I added a driver which creates new instance of PQN and maps Proximal operators to Projection operators..I can handle L1 and probability simplex as well...Will update the code soon...Note that a variant of PQN can do proximal quasi newton but it's still research-y...Several recent papers focused on it http://arxiv.org/pdf/1206.1623v13

Basically with PQN as the solver running on each worker, ADMM will do distributed/multi-core consensus on top of it...For separable functions (like logistic) it will look like this:
minimize F1(x1) + F2(x2) + ...Fn(xn)
s.t x1 = z, x2 = z, ..., xn = z
Here F1 is composite function which satisfies both f(x) and constraint g(z).

From Spark's perspective the master only keeps one vector z and all the solver memory is moved to workers. With 16 GB RAM, we can still do logistic with 2 x 10^9 x 8 byte = 2B features with theoretical guarantees.
 
For non-separable function (neural net) it will be more fun...we will come to it after handling separable functions.
The baseline will be block coordinate descent with PQN on each block until we come up with something better (Gauss Sidel Iterations)
http://opt.kyb.tuebingen.mpg.de/papers/opt2013_submission_1.pdf 

Of course in breeze we are not allowed to use RDD, but does it make sense to define an interface for ADMM based consensus solvers ? We can discuss in person as well..

dlwh: Why not an interface like:

trait SeparableDiffFunction[T] {
  def apply(x: T): IndexedSeq[(Double, T)]
}

?

maybe with a method to turn it into a normal DiffFunction given an OpAdd
for T?

debasish83: Let me add the experimental version with RDD in it and you can help defining the clean interface...I feel we will need different interfaces for separable and non-separable functions.

debasish83: I added the PQN driver and the projection operators for L1(x) <= c and ProbabilitySimplex...Somehow in my experiments I am unable to reproduce the results from Figure 1 of PQN paper...OWLQN is first run to extract the parameter lambda*L1(x*) to generate the constraint L1(x) <= c for PQN

runMain breeze.optimize.proximal.NonlinearMinimizer 500 0.4 0.99

Elastic Net with L1 and L2 regularization. 

Linear Regression:
Issues:
1. Max Iter-ed at 500
2. Worse than OWLQN and naive ADMM in Objective

owlqn 678.072 ms iters 173
pqnSparseTime 30600.365 ms iters 500
owlqnObj -145.38669700980395 pqnSparseObj -135.15057488775597

Logistic Regression:
Cons: 
1. Runtime higher than OWLQN
2. Worse Objective than OWLQN and naive ADMM

owlqn 187.894 ms iters 74 pqn 758.073 ms iters 28
objective owlqnLogistic 52.32713379333781 pqnLogistic 81.37098398138012 

I am debugging the code further but any pointers will be great. I don't think this is the expected behavior from PQN on L1/ProbabilitySimplex constraint as per paper.

dlwh: Random question: does it seem to be line searching a lot (relative to
OWLQN, or runs with a box constraint?)

debasish83: I think my L1 projection has bugs in it...L1(x) is built using ProbabilitySimplex and PQN is beating naive ADMM and at par with QuadraticMinimizer for ProbabilitySimplex.

I will take a closer look into L1 projection for bugs but for ADMM based multicore/distributed consensus I will most likely choose OWLQN for elastic net and PQN for other constraints.

Here are the results with ProbabilitySimplex:

minimize f(x) s.t 1'x = c, x >= 0, c = 1

Linear Regression with ProbabilitySimplex, f(x) = ||Ax - b||^2

Objective pqn 85.34613832959954 nl 84.95320179604967 qp 85.33114863196366
Constraint pqn 0.9999999999999997 nl 1.001707509961072 qp 1.000000000000004
time pqn 150.552 ms nl 15847.125 ms qp 96.105 ms

Logistic Regression with ProbabilitySimplex, f(x) logistic loss from mllib
 
Objective pqn 257.6058563777358 nl 257.4025971846134
Constraint pqn 0.9999999999999998 nl 1.0007230450802203
time pqn 94.19 ms nl 25160.317 ms

debasish83:
@mengxr I refactored ADMM based ConsensusMinimizer to mllib.optimization since it needs to do RDD operations. More results on that will be added in Spark PR. It will use NonlinearMinimizer as a solver that runs on each worker.

NonlinearMinimizer right now supports all the projection operators through PQN and ProximalL1() through OWLQN. I will send a seperate email to John so that he can review the ProjectL1() operators that I added.

Later the plan is to merge PQN and OWLQN to Proximal Quasi Newton method. I will open up the PR for reviews with @yuekai (author of the Matlab code https://github.com/yuekai/pnopt that I adapted) 

Here are some benchmark results and comparison of breeze.optimize.proximal.NonlinearMinimizer with breeze.optimize.proximal.QuadraticMinimizer solver for the respective constraints:

runMain breeze.optimize.proximal.NonlinearMinimizer 400 0.4 0.99

Linear Regression with Elastic Net formulation:

owlqn 2699.234 ms iters 825 sparseQp 359.34 ms iters 68
nlSparseTime 3989.946 ms iters 997
owlqnObj -111.47997989802914 sparseQpObj -111.47977512157428 nlSparseObj -111.47997989810649

Linear Regression with Bounds
qpBox 36.74 ms iters 78
pqnBox 2111.375 ms iters 76
qpBoxObj -59.97519548154472 nlBoxObj -59.96772680436779

Logistic Regression with Bounds
Objective pqn 116.83573241482054 time 1325.132 ms

Linear Regression with ProbabilitySimplex
Objective pqn 95.04618308050102 qp 95.03032612382239
Constraint pqn 0.9999999999999994 qp 1.0000000000000084
time pqn 221.27 ms qp 123.771 ms

Logistic Regression with ProbabilitySimplex
Objective pqn 257.6652169115999
Constraint pqn 0.9999999999999998
time pqn 106.671 ms

You can see since ranks are low, QuadraticMinimizer convergence is much faster !

I am adding the test-cases and then @dlwh please take a pass at it.

dlwh:
very nice! I'll take a look later this evening

debasish83:
I am testing it for the first time but PQN with Identity projection (basically BFGS) shows lot of failures like this:

ERROR ProjectedQuasiNewton Failure! Resetting history: breeze.optimize.LineSearchFailed: Grad norm: 1.0023 Dir Norm: 0.4430
ERROR ProjectedQuasiNewton Failure! Resetting history: breeze.optimize.LineSearchFailed: Grad norm: 0.5606 Dir Norm: 0.2766
ERROR ProjectedQuasiNewton Failure! Resetting history: breeze.optimize.LineSearchFailed: Grad norm: 0.3879 Dir Norm: 0.7122
ERROR ProjectedQuasiNewton Failure! Resetting history: breeze.optimize.LineSearchFailed: Grad norm: 0.8666 Dir Norm: 0.4404

Have you tested identity projection ? It should give us BFGS result

debasish83:
I added linear regression focused tests but I wanted to add some more tests on logistic regression + L1 /probability simplex...

debasish83:
Also the PQN bug with Identity constraints needs to be fixed...The first test captures it:

ERROR ProjectedQuasiNewton Failure! Resetting history: breeze.optimize.LineSearchFailed: Grad norm: 0.6642 Dir Norm: 0.6810
ERROR ProjectedQuasiNewton Failure! Resetting history: breeze.optimize.LineSearchFailed: Grad norm: 1.4756 Dir Norm: 0.4503
ERROR ProjectedQuasiNewton Failure! Resetting history: breeze.optimize.LineSearchFailed: Grad norm: 0.4615 Dir Norm: 3.1684
ERROR ProjectedQuasiNewton Failure! Resetting history: breeze.optimize.LineSearchFailed: Grad norm: 0.1435 Dir Norm: 4.3171
ERROR ProjectedQuasiNewton Failure! Resetting history: breeze.optimize.LineSearchFailed: Grad norm: 0.1209 Dir Norm: 5.0028
ERROR ProjectedQuasiNewton Failure! Resetting history: breeze.optimize.LineSearchFailed: Grad norm: 0.0292 Dir Norm: 4.7222
ERROR ProjectedQuasiNewton Failure! Resetting history: breeze.optimize.LineSearchFailed: Grad norm: 0.0056 Dir Norm: 2.5246
ERROR ProjectedQuasiNewton Failure! Resetting history: breeze.optimize.LineSearchFailed: Grad norm: 0.0062 Dir Norm: 4.8080
ERROR ProjectedQuasiNewton Failure! Resetting history: breeze.optimize.LineSearchFailed: Grad norm: 0.0008 Dir Norm: 3.8321
ERROR ProjectedQuasiNewton Failure! Resetting history: breeze.optimize.LineSearchFailed: Grad norm: 0.0027 Dir Norm: 4.7936
ERROR ProjectedQuasiNewton Failure! Resetting history: breeze.optimize.LineSearchFailed: Grad norm: 0.0008 Dir Norm: 3.5891
ERROR ProjectedQuasiNewton Failure! Resetting history: breeze.optimize.LineSearchFailed: Grad norm: 0.0006 Dir Norm: 3.6643

[info] NonlinearMinimizerTest:
[info] - Nonlinear Minimization with Identity constraint compared to BFGS (924 milliseconds)
[info] - Nonlinear Minimization with positivity constraint compared to Octave (60 milliseconds)
[info] - Nonlinear Minimization with bounds constraint compared to QuadraticMinimizer (29 milliseconds)
[info] - Nonlinear Minimization with L1 constraint compared to Octave (189 milliseconds)
[info] - Nonlinear Minimization with probability simplex compared to Octave (36 milliseconds)

dlwh:
is that with INFO logging suppressed?

debasish83:
I did not turn on any login...u can reproduce the behavior in pqn with identity as well

dlwh:
ok, so probably it's not logging info. I was hoping it was erroring and
restarting repeatedly, because that would be easy to deal with, but alas.

debasish83:
I will take a closer look later..did any user complain abt it ?

dlwh:
nope. You're the first. I'm looking into it too (if you're willing, I'd
definitely appreciate the extra pair of eyes!) This code is pretty new to
me, too, so I'm having to poke around.

dlwh:
I think the SPG algorithm might just be totally wrong. It wasn't tested
separately and it's failing on some simple problems.

dlwh:
yeah, it's definitely wrong.

debasish83:
Nice...That might explain why l1(x) < x was popping...I will also map it to paper...how about compact hessian ?

debasish83:
I will ask professor nocedal how to test B matrices...my eigen check did not work

dlwh:
alright, I'm giving up for the evening. what I have is here:
https://github.com/dlwh/breeze/tree/fix_pqn

I don't know what's going on. I now have something that doesn't get line
search errors, but it's still much slower than BFGS.

I also get test failures in the PQN test and the (new) PSGD test. Not
really sure. Seems like it's getting stuck in the flat parts of the
functions and can't quite close the gap to the optimum?

This may also be of use:
https://github.com/levinboimtomer/SPG/blob/master/SPG.py

I see a few differences (mostly in the line search), but otherwise it seems
ok?

debasish83:
I will look at ur changes tomorrow

debasish83:
If projection is identity why are we doing non-monotone line search in SPG ?

debasish83:
override protected def initialHistory(f: DiffFunction[T], init: T): History = {
    History(0.1, IndexedSeq.empty)
  }
alphaBB = 0.1 is not used anywhere in the algorithm right ? It will be overwritten by first BBStep

dlwh:
It's used in the first iteration. first line of loop in Algorithm 2 of
http://jmlr.csail.mit.edu/proceedings/papers/v5/schmidt09a/schmidt09a.pdf

I don't know what you're asking wrt line search.

debasish83:
SPG keeps the history of fvals and takes the max...they called it nonmonotone in paper...

debasish83:
B matrix construction might have issue as well...rest of ur fixes look good to me

dlwh:
ah, yeah, that history length is configurable via the constructor (fMemory,
probably needs a better name). But I would think that in theory it should
still converge?

debasish83:
In your SPG test the last one with H and q makes sense to me but except that do you think SPG can solve the complicated functions whether the gram/hessian matrix can change...if that's the case then why do we need PQN paper...just run SPG on its own...

debasish83:
I am looking closely into the complicated function test inside PQNTest...that has to pass compared to BFGS...if not there are bugs in B matrix construction or some parameter tuning...

debasish83:
CompactHessian : val sigma = y.dot(y) / yTs
Paper page 3 says sigma is typically yTs/y.dot(y)...Is sigma used in a different way in CompactHessian...

debasish83:
Also this complicated function, could you please add it to LBFGSTest as well...I added and I am getting:
ERROR LBFGS Failure! Resetting history: breeze.optimize.FirstOrderException: Line search zoom failed
The result was correct though

debasish83:
I will add additional SPG tests since most of the tests that I added for QuadraticMinimizer should work fine with SPG except the case that I can handle both the affine constraint Aeq x = beq and bounds lb <= x <= ub in QuadraticMinimizer (the reason why we have wsH) while SPG can't handle it in it's current form...

SPG should work really well for things like large rank NNLS, bounded LS and other projections that we added...

dlwh:
Sure. The function is kind of a worst-case type function for newton-like minimizers backed by line search. It's extremely flat around the optimum, so it's hard to get sufficient descent.

dlwh:
graph

debasish83:
So I read the original 94 paper by Bryd et al REPRESENTATIONS OF QUASI􏰍NEWTON AND THEIR USE IN LIMITED MEMORY METHODS....

There is something fishy...The BFGS update formula wants: 
[Sk'B0Sk Lk]
[Lk'         -Dk]
Now B0 in the code is sigma =  y.dot(s)/y.dot(y)...

But in * operator, v * sigma - u...here sigma should be y.dot(y)/y.dot(s) as per the paper...so the sigma handling does not look quite right...can you also take a closer look ?

Another related optimization is to clean up the LU solve in * operator with Cholesky solve as mentioned in Bryd's paper...I will fix it later

dlwh:
I'll take a look.

I noticed as I was scanning this literature that there's not a lot of consistency in how sigma is defined (i.e. whether or not it's y.dot(s)/y.dot(y) or its reciprocal). In the various papers, they usually get it "right," flipping sigma as necessary.

debasish83:
ok...other than that inconsistency, the rest of the CompactHessian looks good to me as per the paper...should I merge your branch to my PR or you will add the changes to master ?

debasish83:
BFGS is doing strong wolfe but PQN is doing bisection...why not strong wolfe line search in PQN ?

debasish83:
I updated the line search to strong wolfe and I do not see the failures now...SPG anyway I won't test anything other than quadratic as the line search over there do not have strong properties as strong wolfe...

debasish83:
With Strong Wolfe I see occasionally zoom failure:
ERROR ProjectedQuasiNewton Failure! Resetting history: breeze.optimize.FirstOrderException: Line search zoom failed
But this error shows up in LBFGS as well if you add the complicated function to the test.
The tests pass though

debasish83:
The remaining mystery is why L1 projection is not working....I sent email to Mark and I am looking into his matlab implementation

debasish83:
I added a test for ProjectionL1 and it works fine with QuadraticMinimizer...so the ProjectionL1 based on Duchi's paper is correct...

I added the test in NonlinearMinimizer and seems like almost all innerOptimizer are maxIter at 50 which makes it snow but the results were correct....I am now testing it in SPG...

debasish83:
ok...there is a difference in line search...in SPG Mark used cubic interpolation but we are using bisection...should we use cubic in SPG as well ?

dlwh:
sure. Could you code that up?

debasish83:
It will be a modification to StrongWolfeLineSearch right? Looks like we can send fval to the constructor of StrongWolfeLineSearch...

dlwh:
Sure

debasish83:
Well StrongWolfe has this zoom thingie which perhaps we don't need in PQN...Mark's paper says Cubic interpolation but did not say anything about StrongWolfe...

dlwh:
you can always try and see how it goes :)

debasish83:
I am confused...In the paper this is the formula for BB update:

xk+1 ← xk + αdk
sk ← xk+1 − xk
yk ← ∇qk(xk+1) − ∇qk(xk)
αbb ←ykTyk/sTkyk

In PQN Matlab code from Mark's website:

        y = g-g_old;
        s = x-x_old;
        if bbType == 1
            alpha = (s'*s)/(s'*y);
        else
            alpha = (s'*y)/(y'*y);
        end

bbType default is 1...

Which formula is correct ?

dlwh:
no idea :)

Whichever one works better, I guess.

debasish83:
In the matlab code, B matrix construction is same as what we have...For BB, a different formula is used but the one we have works as well...
Right now for L1(x) <= c projection, max iter is the issue...I don't think line search is a big deal here since after all the max iter we are still passing the test...
In Mark's matlab implementation, I see there are two hard termination criterias other than the usual criterias:
1. abs(g'direction) < 1e-6: directional derivative too small
2. if line search fails, the code uses lambda = 0.0 and does not reset history the way we do:
if(alpha < alphaMin) {
      println(s"line search failed")
      throw new LineSearchFailed(normGradInDir, norm(direction))
}
Do we have one of the convergence criteria as abs(g'direction) ?

debasish83:
should I checkin the slow test ? You can take a look at it as well

dlwh:
We don't have abs(g'direction). We could add it. It seems kind of dangerous, and seems like it calls for a history reset rather than a stop. The proper way to stop, as I understand it, in constrained optimization is if norm(proj(x - g) - x), inf) is small. 

Lambda = 0 means no step?

dlwh:
which test is the slow test? how slow?

debasish83:
I added a test for L1(x) <= c...the answer it comes up with is correct but it takes ~100 outer iters but each inner iter maxiters at 50 which makes the test really slow...

debasish83:
Let's focus on SPG first. The BB formula is now correct, I cleanup few things (like normGradInDir sign modification) and added the curvilinear flag which will be useful when I migrate PQN to Proximal Quasi Newton. 

Idea here is that if curvilinear is false don't do the projection inside line search but find descent direction through projection (exactly as the paper). For such cases StrongWolfe line search can be used. 

If curvilinear is true, then we do the projection inside line search and descent direction finding does not do projection (the version that you added). With bisection based line search that makes the convergence for ProjectL1 extremely slow.

> testOnly breeze.optimize.SpectralProjectedGradientTest

ERROR SpectralProjectedGradient Failure! Resetting history: breeze.optimize.FirstOrderException: Line search zoom failed
ERROR SpectralProjectedGradient Failure! Resetting history: breeze.optimize.FirstOrderException: Line search zoom failed
ERROR SpectralProjectedGradient Failure! Resetting history: breeze.optimize.FirstOrderException: Line search zoom failed
ERROR SpectralProjectedGradient Failure! Resetting history: breeze.optimize.FirstOrderException: Line search zoom failed
ERROR SpectralProjectedGradient Failure! Resetting history: breeze.optimize.FirstOrderException: Line search zoom failed
ERROR SpectralProjectedGradient Failure! Resetting history: breeze.optimize.FirstOrderException: Line search zoom failed
ERROR SpectralProjectedGradient Failure! Resetting history: breeze.optimize.FirstOrderException: Line search zoom failed
ERROR SpectralProjectedGradient Failure again! Giving up and returning. Maybe the objective is just poorly behaved?
[info] SpectralProjectedGradientTest:
[info] - optimize a simple multivariate gaussian (696 milliseconds)
[info] - optimize a simple multivariate gaussian with projection (447 milliseconds)
[info] - simple linear solve without projection (444 milliseconds)
[info] - simple linear solve with L1 projection (368 milliseconds) 

Runtime is in check but I am getting Line search zoom exception from Strong Wolfe. I am adding a cubic line search as per Mark's matlab reference. May be due to non-monotonicity in BB Step, Strong Wolfe is not applicable here. Mark told me Strong Wolfe needs to be carefully modified if we want to use it in SPG.

If you run the test with curvilinear=true, we go back to your fix_pqn version but then you will see ProjectL1 does not converges to correct solution.

NonlinearMinimizerTest passes as well but we get lot of "line search zoom failed" as well. So it has definitely something to do with optimization in Strong Wolfe that is only valid for Smooth QuasiNewton optimization.

debasish83:
I added Backtracking line search as default for both PQN and SPG. I can add the cubic interpolation based line search but definitely zoom idea from StrongWolfe does not help.

I reran the benchmarks:

ElasticNet Formulation

Linear Regression

owlqn 5754.259 ms iters 1621 sparseQp 355.579 ms iters 78
nlSparseTime 249588.568 ms iters 243
owlqnObj -110.11580326027442 sparseQpObj -110.11563234413366 nlSparseObj -110.1158032482145

nlSparse here is using ProjectionL1 and we compare it with OWLQN. OWLQN is faster.

Linear Regression with Bounds

qpBox 35.651 ms iters 85
nlBox 1304.603 ms iters 46
qpBoxObj -59.057521023294704 nlBoxObj -59.050115208103925

Logistic Regression with Bounds

Objective nl 124.85981340462786 time 380.22 ms

Linear Regression with ProbabilitySimplex

Objective nl 90.96152371872729 qp 90.94486291900262
Constraint nl 0.9999999999999999 qp 0.9999999999999858
time nl 147.435 ms qp 98.727 ms

Logistic Regression with ProbabilitySimplex

Objective nl 259.2691298934092
Constraint nl 1.0
time nl 113.53 ms

Logistic Regression with ProximalL1 and ProjectL1

Objective proximalL1 11.702710034565527 projectL1 11.67664266228523
time proximalL1 84.034 ms projectL1 1656.008 ms

Again with logistic loss we compare proximalL1 through OWLQN and projectL1 through PQN. OWLQN is much faster.

I will ask Mark if he thinks BacktrackingLineSearch is good for this or I should implement cubic interpolation. Could you please take a pass ? It looks good for me for mllib large scale testing now...

This PR will decrease bring PQN closer to OWLQN by using soft-thresholding formula inside PQN's SPG solver: https://github.com/scalanlp/breeze/pull/372

debasish83:
Owlqn and pqn both uses backtracking line search now...I don't think cubic ls will improve projectl1 runtime in pqn

debasish83:
I looked at the iteration count from outerIter and innerIters:
val spgResult = innerOptimizer.minimizeAndReturnState(new CachedDiffFunction(subprob), x)
println(s"outerIter $iter innerIters ${spgResult.iter}")

Seems like most of the innerIters are still at 50.

debasish83:
I also updated backtracking line search to take a initial function value so that we can support SPG's max f...I forgot to add that in my last iter...

For the test with projectL1 we are doing 50 outer iterations (which is good) but almost all innerIters are maxing at 50...so there is a missing termination criteria in FirstOrderMinimizer...

From the paper and Matlab code seems like the SPG optimality condition is ON and does the following:
optTol = 1e-6
optCond = sum(abs(projection(x-g)-x));
% Check optimality
if optCond < optTol
 if verbose >= 1
     fprintf('First-Order Optimality Conditions Below optTol\n');
     break 
Matlab reference does 1 norm on projection(x-g) - x while the Python reference does infinity norm

Basically for SPG, adjustedGradient has to be implemented that can capture the above termination criteria through the following condition in FirstOrderMinimizer:

else if (norm(adjustedGradient) <= math.max(tolerance * adjustedValue.abs, 1E-8))
        Some(FirstOrderMinimizer.GradientConverged)

The paper wants norm(adjustedGradient, 1) <= tolerance but the one we have in Breeze is slightly different...David did this come from OWLQN paper ? Could you please point a reference ?

debasish83:
Also from the original SPG, this code was suppose to give me Projection(x-g) - x in adjustedGrad but I am not sure what it does...The code projectedVector(newX, - newGrad) I doubt gives me Projection(x-g) - x.. it looks very weird to me....also it does another wasteful Projection...

override protected def adjust(newX: T, newGrad: T, newVal: Double):(Double,T) = (newVal,-projectedVector(newX, - newGrad))

debasish83:
Also this use of adjust is not what you intended over here:
 * @param adjustedValue  f(x) + r(x), where r is any regularization added to the objective. For LBFGS, this is f(x).
   * @param adjustedGradient f'(x) + r'(x), where r is any regularization added to the objective. For LBFGS, this is f'(x).

I am not going to use adjustedGradient and adjustedValue in SPG / PQN....It is meant to add smooth regularizers (basically for which there is no need to add projection/proximal L2 is a great example)...

I am going to add another convergence condition that depends on (x-new - state.x) and we maintain this in algorithm state as well...Only for Identity projection this convergence condition is equivalent to the current GradientConverged...

Please let me know your thoughts..

debasish83:
Seems you are also using adjust inside PQN...so if we stick by that notation in SPG we need to add adjust as well (assuming for PQN and SPG adjust is calculating Projection(x-g) - x, there is no test for that. I will add one)

We should let both PQN and SPG converge on FirstOrderMinimizer.GradientConverged but here I am not clear why we have this formula:
norm(adjustedGradient) <= math.max(tolerance * adjustedValue.abs, 1E-8)

I added the L1 projection test in SPG and that fails if I use the above formula but math.max(tolerance * adjustedValue.abs)  but passes if I use norm(adjustedGradient) <= tolerance.

If we assume adjustedGradient has (Projection(x-g) - x) then this is exactly same as Mark's Matlab reference

debasish83:
David, please take a pass at it. The following issues needs to reviewed:

1. Is PQN adjust is correctly generating (Projection(x - g) - x) for FirstOrderMinimizer GradientConvergence

2. Do we need to use adjust in SPG ? Right now it is not used

3. Should we converge using norm(adjustedGradient) <= math.max(tolerance * adjustedValue.abs, 1E-8) or norm(adjustedGradient) <= tolerance

ProjectL1 runtime is still an issue but Mark said it's better to use OWL variants for L1 and that's what I have made default in NonlinearMinimizer.

dlwh:
I have a few minutes now, so I'll start going through.

1) I'll look.
2) I don't think so, like you say, it's for integrated regularizers in the duchi or owlqn style.
3) probably it should be norm(adjustedGradient) <= math.max(tolerance * adjustedValue.abs, tolerance)  ? We might need to override the convergence check in SPG.

math/src/main/scala/breeze/optimize/proximal/NonlinearMinimizer.scala
```diff
@@ -0,0 +1,206 @@
...
+  case class Projection(proximal: Proximal) {
+    def project(x: DenseVector[Double]) : DenseVector[Double] = {
+      proximal.prox(x)
+      x
+    }
+  }
+
+  def apply(proximal: Proximal, maxIter: Int = -1, m: Int = 10, tolerance: Double = 1e-6): FirstOrderMinimizer[BDV, DiffFunction[BDV]] = {
+    proximal match {
+      case ProximalL1() => new OWLQN[Int, DenseVector[Double]](maxIter, m, proximal.asInstanceOf[ProximalL1].lambda, tolerance)
```
dlwh: why the cast here?

math/src/main/scala/breeze/optimize/LBFGS.scala
dlwh: just checking on this again. Why not make this a method on LBFGS.ApproximateInverseHessian directly?

math/src/main/scala/breeze/optimize/SpectralProjectedGradient.scala
```diff
@@ -1,84 +1,131 @@
...
-  minImprovementWindow: Int = 10,
-  val alphaMax: Double = 1e10,
-  val alphaMin: Double = 1e-10,
-  maxIter: Int = 500,
-  val testOpt: Boolean = true,
+  suffDec: Double = 1e-4,
+  minImprovementWindow: Int = 30,
+  alphaMax: Double = 1e10,
+  alphaMin: Double = 1e-10,
+  bbMemory: Int = 10,
```
dlwh: could you add an `@param` for this?

dlwh: nvm.

math/src/main/scala/breeze/optimize/ProjectedQuasiNewton.scala
```diff
@@ -101,72 +116,53 @@ class ProjectedQuasiNewton(tolerance: Double = 1e-6,
...
+   * @param state the current state
+   * @param f The objective
+   * @param dir The step direction
+   * @return stepSize
+   */
+  protected def determineStepSize(state: State, f: DiffFunction[DenseVector[Double]], dir: DenseVector[Double]) = {
+    val x = state.x
+    val grad = state.grad
+
+    val ff = LineSearch.functionFromSearchDirection(f, x, dir)
```
dlwh: do we not need to project inside the line search?

dlwh:
re: 1), I don't know. maybe best to just override convergence check?

math/src/main/scala/breeze/optimize/ProjectedQuasiNewton.scala
debasish83: Matlab code and the paper did not project inside PQN line search...

dlwh: ah. Right, alpha is probably always <= 1, so it's safe.

math/src/main/scala/breeze/optimize/linear/PowerMethod.scala
```diff
@@ -0,0 +1,83 @@
...
+ */
+class PowerMethod[T, M](maxIters: Int = 10,tolerance: Double = 1E-5)
+                       (implicit space: MutableInnerProductModule[T, Double], mult: OpMulMatrix.Impl2[M, T, T]) extends SerializableLogging {
+
+  import space._
+
+  case class State private[PowerMethod] (eigenValue: Double, eigenVector: T, iter: Int, converged: Boolean)
+
+  //memory allocation for the eigen vector result
+  def normalize(y: T) : T = {
```
dlwh: could you open a bug against normalize for this, just so it doesn't get lost

math/src/main/scala/breeze/optimize/proximal/NonlinearMinimizer.scala
dlwh: I still don't understand the comment. If f(x) can be a convex function, or a non-linear function, then can't f literally be any function? since linear functions are a subset of convex functions...

math/src/main/scala/breeze/optimize/proximal/Proximal.scala
```diff
@@ -0,0 +1,238 @@
...
+  def prox(x: DenseVector[Double], rho: Double = 0.0) = {
+    val atx = at * x
+    val anorm = norm(a, 2)
+    val scale = (b - atx) / (anorm * anorm)
+    val ascaled = a * scale
+    x += ascaled
+  }
+}
+
+case class ProximalL1() extends Proximal {
```
dlwh: why not just have lambda be a constructor param here?
```scala
case class ProximalL1(lambda: Double) {
  def this() = this(1.0)

// ...
}
```

dlwh:
ok, I'm happy if the comments I just made are addressed and we decide what to do about convergence. (That decision can be to punt on it for now and revisit when we actually know what to do.)

math/src/main/scala/breeze/optimize/proximal/Proximal.scala
debasish83: same issue...For sparse coding/plsa each user/document has a different regularization based on how popular the user/document is based on the number of ratings/counts....so during those 100K solves we just need to update the lambda in one object through setLambda...

math/src/main/scala/breeze/optimize/proximal/NonlinearMinimizer.scala
debasish83: Yes...I cleaned it...TODO here is to experiment with TRON for non-convex functions like sigmoid activation and extend PQN to Proximal operators...

math/src/main/scala/breeze/optimize/LBFGS.scala
debasish83: Yeah I will clean it...I will also add similar method on ProjectedQuasiNewton.CompactHessian...seems like they will be useful in Block Coordinate Descent ideas...

math/src/main/scala/breeze/optimize/linear/PowerMethod.scala
debasish83: opened up an issue for it

debasish83:
I am still confused on this:

if (norm(adjustedGradient) <= math.max(tolerance * adjustedValue.abs, 1E-8))
   Some(FirstOrderMinimizer.GradientConverged)

I have used convergence criterias like norm(adjustedGradient) <= tolerance * sqrt(problemSize) but I am not clear why we have adjustedValue over here...

Can you please point me to a reference related to the convergence condition ?

Both SPG and PQN papers needs norm(adjustedGradient) <= tolerance as a convergence condition where adjustedGradient is Projection(x-g) - x

For cases where projection is identity this formula becomes norm(g) <= tolerance which should be consistent with BFGS as well...

I am not sure about OWLQN because when I use norm(g) <= tolerance in my testing, OWLQN gets lot slower...Is it a OWLQN specific optimization ? If it is then it should be defined inside OWLQN and not FirstOrderMinimizer...

dlwh:
I set it up that way for scale invariance. 2 * f(x) and f(x) should
converge to the same solution.

No reference. I just made stuff up.

As I've said before here and elsewhere, happy to take PRs to standardize
this.

math/src/main/scala/breeze/optimize/proximal/Proximal.scala
debasish83: I fixed it...I defined it as var lambda: Double and setLambda modifies it...

math/src/main/scala/breeze/optimize/proximal/NonlinearMinimizer.scala
debasish83: cleaned it up...

debasish83:
ok...Let's go with the current convergence and we will rethink on it later. I cleaned OWLQN from NonlinearMinimizer since we feel ADMM/FISTA will converge at par with OWLQN with the tuning. 

I added a generic proximal solver using ADMM. These are for cases where projection does not converge fast enough (L1 for example) or for cases where proximal operators are defined like huber loss for example. Right now ADMM is still not at par in runtime with OWLQN but it's not that bad either.

Some interesting benchmark numbers from ADMM based proximal solver:

> runMain breeze.optimize.proximal.NonlinearMinimizer 400 0.4 0.99

ElasticNet Formulation

Linear regression

owlqn 7707.722 ms iters 1510 sparseQp 372.883 ms iters 60
nlSparseTime 12894.557 ms iters 1286
nlProxTime 15820.851 ms iters 851
owlqnObj -108.01874244271892 sparseQpObj -108.01852458810231 nlSparseObj -108.01869767843407 nlProxObj -108.01873166329395

owlqn 2X faster than nlSparse (uses SPG by default, L1(x) <= s) and nlProx (uses L1 soft-thresholding)

Logistic Regression with ProximalL1 and ProjectL1
Objective proximalL1 13.320283231557987 projectL1 13.319635853660154 owlqn 13.319731554801983
time proximalL1 789.501 ms projectL1 1264.812 ms owlqn 200.971 ms

owlqn 3x faster than proximalL1 (soft thresholding) and projectL1 (L1(x) <= s)

Logistic Regression with ProbabilitySimplex

Objective nl 258.1749325226205 admm 258.174932597232
Constraint nl 0.9999999999999996 admm 0.9999999999999805
time nlProjection 416.15 ms nlProx 9275.012 ms

For probability simplex projection works much nice

I added a TO DO that I will implement FISTA and compare against ADMM since I believe FISTA is currently one of the leading algorithms for smooth f(x) and non-smooth g(z) like L1. I will do this at a later PR. Right now with this we are covered on most of the mllib needs.

We can either merge it now or wait for few weeks till I am done with stress testing this code through large rank factorization with convex loss in Spark.

dlwh:
Great. I'm happy to merge if you could move minEigen from LBFGS.

debasish83:
I will keep it on my private branch...I am not yet sure how useful are they...waiting for reply from Professor Nocedal on validation and tests for ApproximateInverseHessian/CompactHessian

In BFGS we are using strong wolfe...I was wondering that whether we should add an option for line search that does not need new f & g calculation...something like what we did for SPG...thoughts ? It will speed up both mllib classifiers and ADMM based proximal code...I doubt we need to be so accurate in the line search for both the algorithm variants...

dlwh:
sure, sounds good

debasish83:
cleaned up minEigen...the line search is tricky...will open up a PR once I understand it more...Basically I am thinking about some sort of adaGrad style approach to come up with one learning rate number for ADMM line search...it will need more experiments on large datasets

dlwh:
ok, ok to merge?

debasish83:
Can you please take a closer look at the bfgs state manipulation inside NonlinearMinimizer...I am not doing any excessive object allocation thr ?

dlwh:
sorry, where?

Also I'm not quite sure why you're so worried about performance here;
presumably function evaluations are going to be the bottleneck by a wide
margin?

debasish83:
From bfgs I can only extract solution but then users can't see how inner solve is doing...

debasish83:
I agree with the function eval point...I am ok

dlwh:
I see. why not use lbfgs.minimizeAndReturnState? You can cap the iters for LBFGS in its constructor.

debasish83:
Construct a bfgs solver every admm iteration ? That is cleaner but user won't get any info on inner solve which I guess is consistent with pqn...let me do that

debasish83:
Initially I wanted to use the bfgs history from admm iter i in iter I + 1 but that needs more thought

dlwh:
No. You construct the lbfgs once, but you pass in maxIter = maxInnerIter or
whatever, and then you use minimizeAndReturnState

dlwh:
It would be nice to have that functionality in general, but yeah we don't
really support it yet.

debasish83:
Ok will do that

debasish83:
cleaned LBFGS inner iters...it will be great if you can push it to snapshots...I would like to QuadraticMinimizer and NonlinearMinimizer for Spark PRs...once the codebases are stabilized I will send an email for cutting release...If you have an optimization page that I can modify, I would like to add proper documentation...Lot of capabilities are added now through these proximal algorithm (direct and iterative)...It will be great if they can be used in other Apache projects as well...

dlwh:
Right now, basically the only documentation for optimize is here: https://github.com/scalanlp/breeze/wiki/Quickstart#breezeoptimize

That needs to be fixed.

dlwh:
Thanks again for contributing this!

dlwh:
(publishing snapshots now. should be available in 10-15 minutes)

debasish83:
how can I uppdate the documentation ? We will be working on fixing it over the weekend...

dlwh:
feel free to edit the wiki, add new pages, whatever you think is best.

PR #366, opened 2015-02-18
Optimized QR factorization

nikdon:
QR with speed increased by ~20+% for both reduced and complete modes. 

@dlwh could you have a look please?


dlwh:
thanks!

PR #367, opened 2015-02-19
Randomized SVD

nikdon:
SVD of large scale dense matrices that gives significant speed-up. Based on [1, 2].

Any suggestions are welcome

[1][ Finding structure with randomness: Stochastic algorithms for constructing approximate matrix decompositions](http://arxiv.org/abs/arXiv:0909.4061)
[2][ Randomized methods for computing the Singular Value Decomposition (SVD) of very large matrices](http://web.stanford.edu/group/mmds/slides2010/Martinsson.pdf)

### DIFF
--- a/math/src/main/scala/breeze/linalg/functions/svdr.scala
+++ b/math/src/main/scala/breeze/linalg/functions/svdr.scala
@@ -3,23 +3,15 @@ package breeze.linalg.functions
 
 import breeze.generic.UFunc
 import breeze.linalg._
-import breeze.linalg.svd.SVD
+import breeze.linalg.svd.{DenseSVD, SVD}
 import breeze.numerics.{abs, signum}
 import breeze.stats.distributions.Rand
+
 
 /**
  * Approximate truncated randomized SVD
  */
 object svdr extends UFunc {
-
-  case class SVDR[M, V](leftVectors: M, singularValues: V, rightVectors: M) {
-    def U: M = leftVectors
-    def ∑ : V = singularValues
-    def S = ∑
-    def Vt: M = rightVectors
-  }
-
-  type DenseSVD = SVDR[DenseMatrix[Double], DenseVector[Double]]
 
   implicit object SvdR_DM_Impl2 extends Impl2[DenseMatrix[Double], Int, DenseSVD] {
     def apply(M: DenseMatrix[Double], k: Int): DenseSVD =
@@ -77,7 +69,7 @@ object svdr extends UFunc {
 
     val (u, v) = flipSVDSigns(_u, _v)
 
-    SVDR(u(::, 0 until k), _s(0 until k), v(0 until k, ::))
+    SVD(u(::, 0 until k), _s(0 until k), v(0 until k, ::))
   }
 
   /**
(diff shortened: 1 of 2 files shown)


math/src/main/scala/breeze/linalg/functions/svdr.scala
```diff
@@ -0,0 +1,125 @@
...
+import breeze.linalg.svd.SVD
+import breeze.numerics.{abs, signum}
+import breeze.stats.distributions.Rand
+
+/**
+ * Approximate truncated randomized SVD
+ */
+object svdr extends UFunc {
+
+  case class SVDR[M, V](leftVectors: M, singularValues: V, rightVectors: M) {
```
dlwh: we can probably just use the svd.SVD class?

math/src/main/scala/breeze/linalg/functions/svdr.scala
```diff
@@ -0,0 +1,125 @@
...
+import breeze.generic.UFunc
+import breeze.linalg._
+import breeze.linalg.svd.SVD
+import breeze.numerics.{abs, signum}
+import breeze.stats.distributions.Rand
+
+/**
+ * Approximate truncated randomized SVD
+ */
+object svdr extends UFunc {
```
dlwh: at some point I want to figure out the right syntax for typesafe configuration for UFuncs. Something like `svd(M, 'mode := 'random)` but probably not that, exactly.

math/src/main/scala/breeze/linalg/functions/svdr.scala
```diff
@@ -0,0 +1,125 @@
...
+   *
+   * Algorithm 4.3 of "Finding structure with randomness:
+   * Stochastic algorithms for constructing approximate matrix decompositions"
+   * Halko, et al., 2009 (arXiv:909) [[http://arxiv.org/pdf/0909.4061]]
+   */
+  private def randomizedStateFinder(M: DenseMatrix[Double],
+                                    size: Int,
+                                    nIter: Int): DenseMatrix[Double] = {
+    val R = DenseMatrix.rand(M.cols, size, rand = Rand.gaussian)
+    var Y = M * R
```
dlwh: Y doesn't need to be a var

dlwh:
i think it basically looks good. just a few minor comments and a little rampant speculation.

math/src/main/scala/breeze/linalg/functions/svdr.scala
nikdon: absolutely, fixed as well as type

math/src/main/scala/breeze/linalg/functions/svdr.scala
nikdon: fixed

nikdon:
Made cleanup and some refactor. If it still looks good, I think you can merge it.

math/src/main/scala/breeze/linalg/functions/svdr.scala
nikdon: I've thought about that, but code became messy. Something like ```svd.randomized(M, ...)```. But finally decided to make a separate ```svdr```.

dlwh: yeah this doesn't need it, but we have a lot of things that are currently methods because UFuncs can't handle configuration.

dlwh:
awesome. thanks so much for doing this!

PR #368, opened 2015-02-21
Tile UFunc

rakeshchalasani:
Similar to numpy tile
http://docs.scipy.org/doc/numpy/reference/generated/numpy.tile.html

Only implemented for DenseVector and DenseMatrix.
Ex: 
```
tile(DenseVector(1,2),2)  // Creates DenseVector of length 4
tile(DenseVector(1,2),2,2) // Creates (4 x 2) DenseMatrix
tile(DenseMatrix.rand[Double](2,2), 2,3) // Creates (4 x 6) DenseMatrix
```


dlwh:
great thanks!

PR #370, opened 2015-02-21
[WIP] Randomized EVD

nikdon:
Same idea as in #367, but for eigenvalue decomposition of square matrices. Again it's fast for big problems and gives a good approximate result. Some parts of ```svdr``` are used (```randomizedStateFinder```) so it'll be a good idea to move them at another abstraction level and cut the duplicates. If this function is interesting any suggestions are welcome.

### DIFF
--- a/math/src/main/scala/breeze/linalg/functions/evdr.scala
+++ b/math/src/main/scala/breeze/linalg/functions/evdr.scala
@@ -1,0 +1,112 @@
+package breeze.linalg.functions
+
+
+import breeze.generic.UFunc
+import breeze.linalg._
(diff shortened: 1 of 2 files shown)


dlwh:
Thanks! So it seems like this can made to work (via implicit Ops) to work for any type M that has a OpMul.Impl2[M, Dense, Dense] (and possibly OpMul.Impl2[TransM, Dense, Dense]  though that's not strictly necessary?) I.e. It looks pretty easy to make this handle sparse matrices too?

nikdon:
Looks like yes, it's necessary to perform a couple of multiplications

PR #380, opened 2015-03-13
QuadraticMinimizer.updateGram cleaned exceptions

debasish83:
At each iteration of updateGram I was checking if the indices are within bounds but when we call it from mllib ALS, we make sure upfront if the indices are fine. Having conditional checks will affect the performance for example. I cleaned it up and added a comment so that users are aware of this optimization.

### DIFF
--- a/math/src/main/scala/breeze/optimize/proximal/QuadraticMinimizer.scala
+++ b/math/src/main/scala/breeze/optimize/proximal/QuadraticMinimizer.scala
@@ -112,6 +112,8 @@ class QuadraticMinimizer(nGram: Int,
    * of update. No exceptions are thrown explicitly to optimize on the runtime
    */
   def updateGram(row: Int, col: Int, value: Double) {
+    require(row >= 0 && row < n, s"QuadraticMinimizer:updateGram row index out of bounds")
+    require(col >= 0 && col < n, s"QuadraticMinimizer:updateGram col index out of bounds")
     wsH.update(row, col, value)
   }
 
--- a/math/src/main/scala/breeze/optimize/proximal/QuadraticMinimizer.scala
+++ b/math/src/main/scala/breeze/optimize/proximal/QuadraticMinimizer.scala
@@ -107,11 +107,11 @@ class QuadraticMinimizer(nGram: Int,
 
   def getProximal = proximal
 
+  /**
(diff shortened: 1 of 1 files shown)


debasish83:
I integrated NNLS and QuadraticMinimizer to Spark and in my local setup, tests run fine...if we cut a release after merging this PR, I think we should be good for ALS with userConstraint and productConstraint

math/src/main/scala/breeze/optimize/proximal/QuadraticMinimizer.scala
```diff
@@ -107,11 +107,13 @@ class QuadraticMinimizer(nGram: Int,
 
   def getProximal = proximal
 
+  /**
+   * updateGram API is meant to be called iteratively from the user and the user should guarantee correctness
+   * of update. No exceptions are thrown explicitly to optimize on the runtime
+   */
```
dlwh: i mean, require just throws an exception if it fails?

debasish83: Yeah that's true...earlier I wanted to have no checks in updateGram for efficiency...But now I am not sure...This is how I use it from ALS:

val quadraticMinimizer = QuadraticMinimizer(rank, SMOOTH)
    /**
     * Given a triangular matrix in the order of fillXtX above, compute the full symmetric square
     * matrix that it represents, storing it into destMatrix.
     */
    private def fillAtA(triAtA: Array[Double], lambda: Double) {
      var i = 0
      var pos = 0
      var a = 0.0
      while (i < rank) {
        var j = 0
        while (j <= i) {
          a = triAtA(pos)
          if (i == j) {
            quadraticMinimizer.updateGram(i, j, a + lambda)
          } else {
            quadraticMinimizer.updateGram(i, j, a)
            quadraticMinimizer.updateGram(j, i, a)
          }
          pos += 1
          j += 1
        }
        i += 1
      }
    }
So updateGram will be call lot of time...May be it makes sense to not have any checks in it and let user be responsible for correctness ?

dlwh: I think you're prematurely optimizing.

debasish83: hmm...then I guess adding require will be fine...

PR #383, opened 2015-03-17
Index range tests in SparseVector are off by one.

NathanHowell:
There are redundant range tests in SparseVector so the previous behavior was observably correct and had test coverage already. This fix ensures the failures happen at a consistent location.

Example call stack prior to the fix, note that the exception is thrown from `SparseArray.findOffset` and not `SparseVector.apply`:

    Exception in thread "main" java.lang.IndexOutOfBoundsException: Index 12 out of bounds [0,12)
        at breeze.collection.mutable.SparseArray.findOffset(SparseArray.scala:198)
        at breeze.collection.mutable.SparseArray.apply(SparseArray.scala:53)
        at breeze.collection.mutable.SparseArray.apply$mcD$sp(SparseArray.scala:52)
        at breeze.linalg.SparseVector$mcD$sp.apply$mcD$sp(SparseVector.scala:75)
        at breeze.linalg.SparseVector$mcD$sp.apply(SparseVector.scala:73)
        at breeze.linalg.SparseVector$mcD$sp.apply(SparseVector.scala:49)
        at breeze.linalg.TensorLike$class.apply$mcID$sp(Tensor.scala:94)
        at breeze.linalg.SparseVector.apply$mcID$sp(SparseVector.scala:49)

### DIFF
--- a/math/src/main/scala/breeze/linalg/SparseVector.scala
+++ b/math/src/main/scala/breeze/linalg/SparseVector.scala
@@ -71,12 +71,12 @@ class SparseVector[@spec(Double, Int, Float, Long) V](val array: SparseArray[V])
   def contains(i: Int) = array.contains(i)
 
   def apply(i: Int): V = {
-    if(i < 0 || i > size) throw new IndexOutOfBoundsException(i + " not in [0,"+size+")")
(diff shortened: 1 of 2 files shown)


dlwh:
thanks

PR #384, opened 2015-03-19
Convergence a la carte

nightscape:
So here is a first shot at what an API for picking custom convergence criteria could look like.
I'm not expecting this PR to be merged as is, but rather as something to discuss and improve.
Especially I'm not very proud of the type boilerplate (passing [T, History] around every where and even worse the FirstOrderMinimizer[T, DF]#History thing) so maybe someone with more Scala types proficiency can come up with a nicer solution.

### DIFF
--- a/math/src/main/scala/breeze/optimize/FirstOrderMinimizer.scala
+++ b/math/src/main/scala/breeze/optimize/FirstOrderMinimizer.scala
@@ -145,7 +145,9 @@ object FirstOrderMinimizer {
   }
 
   type ConvergenceCheck[T, History] = PartialFunction[State[T, History], ConvergenceReason]
-
+  implicit class RichConvergenceCheck[T, History](val check: ConvergenceCheck[T, History]) extends AnyVal {
+    def ||(otherCheck: ConvergenceCheck[T, History]): ConvergenceCheck[T, History] = check orElse otherCheck
+  }
 
   sealed trait ConvergenceReason {
     def reason: String
@@ -190,16 +192,13 @@ object FirstOrderMinimizer {
       SearchFailed
   }
   def createConvergenceCheck[T, History](maxIter: Int, tolerance: Double, relative: Boolean = true)(implicit space: NormedModule[T, Double]): State[T, History] => Option[ConvergenceReason] =
-    convergedWhen(
-      maxIterationsReached[T, History](maxIter),
-      functionValuesConverged[T, History](tolerance, relative),
-      objectiveNotImproving[T, History](tolerance, relative),
-      gradientConverged[T, History](tolerance, relative),
+    (
+      maxIterationsReached[T, History](maxIter) ||
+      functionValuesConverged[T, History](tolerance, relative) ||
+      objectiveNotImproving[T, History](tolerance, relative) ||
+      gradientConverged[T, History](tolerance, relative) ||
       searchFailed[T, History]
-    )
-  def convergedWhen[T, History](convergenceChecks: ConvergenceCheck[T, History]*): State[T, History] => Option[ConvergenceReason] = {
-    convergenceChecks.reduceLeft(_ orElse _).lift
-  }
+    ).lift
 
   /**
    * OptParams is a Configuration-compatible case class that can be used to select optimization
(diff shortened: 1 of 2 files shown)


nightscape:
See https://github.com/scalanlp/breeze/pull/384/files#diff-ad6864624476e90c0dcd78e32dcdf786R192 on how convergence criteria can be created.

dlwh:
We probably want to add the ability to make checks relative or absolute thresholded. That's been one of the major points of contention. (E.g. the gradient norm check)

nightscape:
Like this?
https://github.com/crealytics/breeze/commit/b662fdb4b2c4272dd666211b63651623f4f9ddca

math/src/main/scala/breeze/optimize/FirstOrderMinimizer.scala
```diff
@@ -184,7 +166,40 @@ object FirstOrderMinimizer {
...
+  }
+  def createConvergenceCheck[T, History](maxIter: Int, tolerance: Double, relative: Boolean = true)(implicit space: NormedModule[T, Double]): State[T, History] => Option[ConvergenceReason] =
+    convergedWhen(
+      maxIterationsReached[T, History](maxIter),
+      functionValuesConverged[T, History](tolerance, relative),
+      objectiveNotImproving[T, History](tolerance, relative),
+      gradientConverged[T, History](tolerance, relative),
+      searchFailed[T, History]
+    )
+  def convergedWhen[T, History](convergenceChecks: ConvergenceCheck[T, History]*): State[T, History] => Option[ConvergenceReason] = {
```
dlwh: maybe name this convergedWhenAny or something? It's not clear if it's a || or a && thing from the name.

nightscape: What do you think of creating an implicit `RichConvergenceCheck` that has `&&` and `||` methods? That way you could arbitrarily combine ConvergenceChecks like Booleans and still get the convergence reason(s) out in the end.

dlwh: sure sounds good.

nightscape: Added a RichConvergenceCheck, which currently only provides `||` concatenation of convergence checks.

nightscape:
Hi David,

sorry for leaving this PR unattended for so long, I've been on parental leave.
Convergence checks can now be specified to be relative and via an `implicit RichConvergenceCheck` they can be concatenated via `||`.
If someone needs `&&` as well we can add that later.
Independent of this PR, personally I would tear out the entire convergence handling from `FirstOrderMinimizer` because it can be seen as an orthogonal aspect to the generation of solution steps.
But this would probably need an API break which should be well-discussed and planned.

So, as a first improvement this PR should at least allow for custom convergence criteria and fine-grained configuration of the existing ones.

nightscape:
Hi David, sorry to bug, but did you have time to check this PR?
Is there anything that you'd like to be improved/added/removed?
Any other concerns or ideas?

dlwh:
sorry, i'll take a look today.

math/src/main/scala/breeze/optimize/ProjectedQuasiNewton.scala
```diff
@@ -78,25 +79,42 @@ class CompactHessian(M: DenseMatrix[Double], Y: RingBuffer[DenseVector[Double]],
...
   type BDV = DenseVector[Double]
-
-  val innerOptimizer = new SpectralProjectedGradient[BDV](
-    tolerance = tolerance,
-    maxIter = 50,
-    bbMemory = 5,
-    initFeas = true,
-    minImprovementWindow = 10,
-    projection = projection
+  def this(tolerance: Double = 1e-6,
```
dlwh: can you fix indentation for this one?

dlwh:
that one thing and then lgtm

nightscape:
Fixed the whitespace issue.
Shall I rebase against latest master and amend the whitespace fixes into the commit where it was messed up?

dlwh:
nah

dlwh:
thanks!

nightscape:
Thanks to you!

PR #386, opened 2015-03-22
Memory Optimization

debasish83:
Memory optimization are done to bring optimize.linear.NNLS runtime closer to mllib NNLS and optimize.proximal.QuadraticMinimizer default close to blas.dposv

NNLS:iterator pattern cleaned for speed, in-place gemv added,initialState API provided for state reuse;PowerMethod:specialized on DenseMatrix and DenseVector for speed;QuadraticMinimizer:iterator pattern cleaned for speed,memory optimization to bring runtime close to dposv


debasish83:
I will take a closer look on comparisons with ml cholesky solver tomorrow...the first iteration is always slow...I am not quite sure why...

debasish83:
By the way sorry to turn the code into a C-style code but I had to make sure no memories are allocated in the whole algorithm except through NNLS.initialize and QuadraticMinimizer.initialize....When you review please let me know if you see any additional memory allocation in algorithm inner loop (iterations) since I am using lot of breeze overloaded functions and might have missed things..

math/src/main/scala/breeze/optimize/linear/NNLS.scala
```diff
@@ -1,32 +1,36 @@
...
   // find the optimal unconstrained step
   private def steplen(ata: BDM, dir: BDV, res: BDV,
               tmp: BDV): Double = {
-    val top = dir.dot(res)
-    tmp := ata * dir
+    val top = dir dot res
+    gemv(1.0, ata, dir, 0.0, tmp)
     // Push the denominator upward very slightly to avoid infinities and silliness
-    top / (tmp.dot(dir) + 1e-20)
+    top / (tmp dot dir + 1e-20)
```
dlwh: i believe this is not what you want. This changes the operation. 

```scala
scala> implicit class Foo(x: Int) { def dot(y: Int) = x * y }
defined class Foo

scala> 3 dot 2 + 3
res1: Int = 15

scala> 3.dot(2) + 3
res2: Int = 9
```

math/src/main/scala/breeze/optimize/linear/NNLS.scala
```diff
@@ -53,32 +57,67 @@ class NNLS(val maxIters: Int = -1) extends SerializableLogging {
...
+   * resetState is set to false, the optimizer will hot start using the previous state. By default resetState is
+   * true and every time reset will be called on the incoming state
+   *
+   * @param ata gram matrix
+   * @param atb linear term
+   * @param initialState initial state for calling the solver from inner loops
+   * @param resetState reset the state based on the flag
+   * @return converged state
+   */
+  def iterations(ata: DenseMatrix[Double],
```
dlwh: this is misnamed now, since it does not return an iterator. why not just fold all of this into minimizeAndReturnState ?

math/src/main/scala/breeze/optimize/proximal/QuadraticMinimizer.scala
```diff
@@ -163,25 +176,55 @@ class QuadraticMinimizer(nGram: Int,
...
+   * initialState. It also exposes alpha and rho control to users who would like to experiment with rho and alpha
+   * parameters of the admm algorithm
+   * @param q linear term for the quadratic optimization
+   * @param rho rho parameter for ADMM algorithm
+   * @param alpha over-relaxation alpha parameter for admm algorithm, default 1.0
+   * @param initialState provide a initialState using initialState API
+   * @param resetState use true if you want to hot start based on the provided state
+   * @return converged state from ADMM algorithm
+   */
+  def iterations(q: DenseVector[Double],
```
dlwh: also misnamed now.

math/src/main/scala/breeze/optimize/linear/NNLS.scala
debasish83: My bad...fixed it

math/src/main/scala/breeze/optimize/linear/NNLS.scala
debasish83: yes that's much clean...did it

debasish83:
any memory specific optimization ? Most of the time I run QuadraticMinimizer first and so hot spot is an issue...I will try the same run after warming JVM tomorrow and finish up the API changes in AM

dlwh:
Remember it's not just that the JVM is warming up: It has to warm up for
each individual class. In general, methods aren't fully optimized until
they're called about 10,000 times. So really you should be running the
algorithm several times, then time it.

-- David

math/src/main/scala/breeze/optimize/proximal/QuadraticMinimizer.scala
debasish83: I cleaned it up...

debasish83:
I added an API to provide upper triangular gram matrix and with that the runtime in the first iteration also dropped...I think QuadraticMinimizer should be able to replace the ML CholeskySolver now...

debasish83:
The first iteration issue is consistent with both NNLS and QuadraticMinimizer...Out of curiosity, I looked at the code and both mllib and Breeze back matrix and vector workspace with Arrays[Double]...so I am really not clear why there is the 2X difference only in initial iterations....Is it due to the overhead from traits that show up in DenseVector/DenseMatrix ? 

During the solve things are clean and so I don't think there are cases where BLAS using native memory is faster than QuadraticMinimizer sending memory to LAPACK to work on...

dlwh:
using a DenseVector for the first time incurs a lot of overhead:  operators
are populated into the multimethod maps, lots of interfaces are loaded,
etc. You really clock each of these after running the exact same code path
at least once. E.g.

def time(x: =>Unit) {
   for(i <- 0 until N) x

   val in = System.currentTimeMillis();
   x;
   out = System.currentTimeMillis();
}

dlwh:
what's the status here? Can I merge this? I really want to release a fix for the SparseVector bug

debasish83:
I am ok with this...moving from DenseVector/DenseMatrix to Array will make the code ugly

debasish83:
Please let me know when you cut 0.11.2...

dlwh:
tongiht

PR #390, opened 2015-03-26
QR for floats

nikdon:
QR for floats with tests. Copy-paste of the implementation for doubles with float types.

Any comments on what can be change are welcome


math/src/main/scala/breeze/linalg/functions/qr.scala
```diff
@@ -177,6 +207,68 @@ object qr extends UFunc {
       (Q(::, 0 until mc), A(0 until mc, ::))
     }
   }
+
+  private def doQr_Float(M: DenseMatrix[Float], skipQ: Boolean)
```
dlwh: i should probably `@expand` stuff like this.

nikdon: Thanks for tip, I'll try to think about that..

PR #392, opened 2015-04-04
QuadraticMinimizer bug fixes

debasish83:
After experimentation inside mllib ALS I found out two bugs

1. QuadraticMinimizer Quasi Definite System was not reinitializing correctly:
[H A']
[A 0]
I did not re-initialize 0 correctly and so when state was re-used results were incorrect. This issue will only show up in EQUALITY formulation. Added a test-case to guard against it. 

2. QuadraticMinimizer used to return the primal variable (x) but changed it to return dual variable (z) which gives a projection style result. The algorithm converges only if x and z are within absTol.
Basically in place of getting numbers like 1e-5 for POSITIVE constraint, now we will get 0.0

3. Improved convergence of ProbabilitySimplex using L1 rho formula


debasish83:
@dlwh give it one week or so...I have started testing the convergence and runtime of NonlinearMinimizer as well inside Spark through Alternating Minimization (Matrix factorization for large ranks and loss functions like loglikelihood and kl divergence)...next week I will find out more convergence related issues from NonlinearMinimizer/SPG based flow as well...

debasish83:
@dlwh you can merge it...I did not get time to test NonlinearMinimizer variants last few weeks...

PR #395, opened 2015-04-16
Implictly determined fitting implementation of LU and softmax. But did n...

bashimao:
...ot use it.

Well this is rather trivial. Most people probably always rely on the built-in implementation. As long as you do so, this worked nice. If (as in my case) you provide your own implementation, the previous code will fail.


dlwh:
do you have evidence it's not being used? the implicit lookup should actually be using the right things.

PR #402, opened 2015-04-28
Fixed "// TODO: use := when that's available" and provided implementatio...

bashimao:
...n for Float type.


dlwh:
could you delete the cast float impl? It's causing the tests to not compile.

bashimao:
Done!  (p.s.: sbt did not complain on my machine)

dlwh:
thanks!

PR #413, opened 2015-06-18
Approximate hessian matrix using central differences

skavulya:
This is an initial implementation of the approximate hessian matrix using the central difference equation. It is the same implementation used by optimHess() in the R stats package. I used the example in R for the initial  test case. I am new to Breeze so I might have used the wrong data structures and conventions. Please let me know what changes I need to make.

### DIFF
--- a/math/src/main/scala/breeze/optimize/SecondOrderFunction.scala
+++ b/math/src/main/scala/breeze/optimize/SecondOrderFunction.scala
@@ -1,6 +1,7 @@
 package breeze.optimize
 
 import breeze.linalg.operators.OpMulMatrix
+import breeze.linalg.support.CanCopy
 import breeze.linalg.{DenseMatrix, DenseVector}
 import breeze.math.{InnerProductVectorSpace, MutableInnerProductVectorSpace, VectorSpace}
 import breeze.stats.distributions.Rand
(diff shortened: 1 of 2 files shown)


math/src/main/scala/breeze/optimize/SecondOrderFunction.scala
```diff
@@ -68,38 +68,85 @@ object SecondOrderFunction {
...
+    // second order differential using central differences
+    for (i <- 0 until n) {
+      x(i) = x(i) + eps
+      val df1 = df.gradientAt(x)
+
+      x(i) = x(i) - 2 * eps
+      val df2 = df.gradientAt(x)
+
+      val grad = (df1 - df2)/(2*eps)
+      for (j <- 0 until n) {
```
dlwh: this loop can just be H(i, ::) := grad

skavulya: Thanks. I made the change, and updated the code to use a copy of the input vector. I was thinking of refactoring my code, and creating CentralDiffFunction class which extends SecondOrderFunction. The hessian calculation would be done in calculate2(). However, I am not sure how to represent the matrix in Breeze. Would it be an IndexedSeq[T]? If so, is there a way I can convert the IndexedSeq to a matrix?

dlwh:
thanks!

math/src/main/scala/breeze/optimize/SecondOrderFunction.scala
dlwh: for now I'd just force T to be DenseVector

skavulya:
Thanks for all your help!

PR #420, opened 2015-07-23
Fixed bug in speed improved CanTransformValues

bashimao:
Full story available at: https://github.com/scalanlp/breeze/commit/a81e0a055c0429d9bbaaee3ca5606673632f71c5

### DIFF
--- a/math/src/main/scala/breeze/linalg/DenseVector.scala
+++ b/math/src/main/scala/breeze/linalg/DenseVector.scala
@@ -407,10 +407,10 @@ object DenseVector extends VectorConstructors[DenseVector]
           }
         } else {
           val end = offset + stride * length
-          var i = offset
-          while (i != end) {
-            d(i) = fn(d(i))
-            i += stride
+          var j = offset
+          while (j != end) {
+            d(j) = fn(d(j))
+            j += stride
           }
 
         }
--- a/math/src/main/scala/breeze/linalg/DenseVector.scala
+++ b/math/src/main/scala/breeze/linalg/DenseVector.scala
@@ -397,26 +397,20 @@ object DenseVector extends VectorConstructors[DenseVector]
     new CanTransformValues[DenseVector[V], V] {
       def transform(from: DenseVector[V], fn: (V) => V) {
         val d = from.data
+        val length = from.length
         val stride = from.stride
 
         val offset = from.offset
         if (stride == 1)  {
-          if (offset == 0) {
-            cforRange(0 until d.length) { j =>
-              d(j) = fn(d(j))
-            }
-          } else {
-            cforRange(offset until offset + from.size) { j =>
-              d(j - offset) = fn(d(j))
-            }
+          cforRange(offset until offset + length) { j =>
+            d(j) = fn(d(j))
           }
         } else {
-          var i = 0
-          var j = offset
-          while(i < from.length) {
-            d(j) = fn(d(j))
-            i += 1
-            j += stride
+          val end = offset + stride * length
+          var i = offset
+          while (i != end) {
+            d(i) = fn(d(i))
+            i += stride
           }
 
         }


dlwh:
Hrm, I seem to be totally wrong. I swore I measured specifically this.


With your patch, everything is fine. (InPlace for DenseVector being the relevant change.)

```
[info]                 MappingUfuncDenseMat  9.03 ================
[info]                 MappingUfuncDenseVec  8.82 ================
[info]       MappingUfuncDenseMatWithStride  9.80 =================
[info]          MappingUfuncDenseVecInPlace  2.45 ====
[info]             MappingUfuncArrayInPlace  2.46 ====
[info]          MappingUfuncDenseMatInPlace  2.43 ====
[info]              MappingUfuncArrayHarder 15.89 ============================
[info]           MappingUfuncDenseVecHarder 15.68 ============================
[info]           MappingUfuncDenseMatHarder 15.40 ============================
[info] MappingUfuncDenseMatHarderWithStride 16.47 ==============================
[info]  MappingUfuncDenseVecHarderMapValues 15.60 ============================
[info]  MappingUfuncDenseMatHarderMapValues 15.73 ============================
[info] MappingUfuncArrayHarderInlineInPlace 10.26 ==================
[info]    MappingUfuncDenseMatHarderInPlace 10.19 ==================
[info]    MappingUfuncDenseVecHarderInPlace 10.22 ==================
[info]       MappingUfuncArrayHarderInPlace 10.84 ===================
```

dlwh:
Hrm, seems that I must have (incorrectly) generalized from the CanMapValues. If I apply the same change to CanMapValues: 

```scala
        if (stride == 1) {
//          if (off == 0) {
//            cforRange(0 until arr.length) { j =>
//              arr(j) = fn(d(j))
//            }
//          } else {
            cforRange(0 until arr.length) { j =>
              arr(j) = fn(d(j + off))
            }
//          }
```

There's a very noticeable slowdown of around 40% (the DenseVectors, but not InPlace):

```
[info]                 MappingUfuncDenseMat  9.73 ===========
[info]                 MappingUfuncDenseVec 10.25 ============
[info]       MappingUfuncDenseMatWithStride 10.03 ============
[info]          MappingUfuncDenseVecInPlace  2.49 ===
[info]             MappingUfuncArrayInPlace  2.54 ===
[info]          MappingUfuncDenseMatInPlace  2.49 ===
[info]              MappingUfuncArrayHarder 16.13 ===================
[info]           MappingUfuncDenseVecHarder 22.87 ============================
[info]           MappingUfuncDenseMatHarder 15.48 ===================
[info] MappingUfuncDenseMatHarderWithStride 17.15 =====================
[info]  MappingUfuncDenseVecHarderMapValues 24.38 ==============================
[info]  MappingUfuncDenseMatHarderMapValues 16.60 ====================
[info] MappingUfuncArrayHarderInlineInPlace 10.37 ============
[info]    MappingUfuncDenseMatHarderInPlace 10.31 ============
[info]    MappingUfuncDenseVecHarderInPlace 10.34 ============
[info]       MappingUfuncArrayHarderInPlace 10.29 ============
```

dlwh:
i guess it probably has something to do with the second array? But the HotSpot wiki says that that shouldn't matter:

https://webcache.googleusercontent.com/search?q=cache:0CDR--9A5KYJ:https://wikis.oracle.com/display/HotSpotInternals/RangeCheckElimination+&cd=1&hl=en&ct=clnk&gl=us

(The last bit about how multiple arrays are ok.)

I dunno. Maybe it generates bytecode that doesn't trigger the matcher?

bashimao:
Yes, I would draw the same conclusion from this description. The amount of arrays accessed should not really affect the outcome as long as we use a simple <code>offset(val) + index(var incremented by scale(val))</code> addressing scheme for each of them. We also have all relevant values loaded in local immutable val-fields. So that should also not matter. The JIT compiler should be able to handle this properly.

Did you by any chance use the "optimise"-flag for compiling the Scala code? There could be a remote chance that the Scala compiler transparently messes up the byte-code. However, it surely is very unlikely.

<strong>Just an idea:</strong>
You could try to to replace the loop (currently <code>offset(val) + index(var incremented by scale(val)</code>) with <code>offset(val) + stride(val) * index(var increment by 1)</code>. If there is a somewhat sophisticated optimizer, it should be able to understand your intention and easily eliminate all range checks because a <code>for (i = 0; i < N; i++)</code> loop is trivial to analyze. If so, the compiler might also be able to automatically replace the multiplication with addition. Most C-Compilers are able to do this. But I am not sure about the Java-JIT. It is definitely a far shot, but maybe worthwhile trying.

bashimao:
Please forget my previously mentioned "idea". I just tested it. The -15% speed when looping with increments and multiplying in the loop. My conclusion, the JIT is either not clever enough to replace the multiplication with additions (more likely) or it its keeping the range checks in that situation (less likely).

bashimao:
@dlwh: Can you also update "DenseVector[T].foreach", please?

dlwh:
I think the latter answer is indeed what's happening. The docs say very clearly that it can only do ABCE for constant strides. What's surprising is that I found you can still get hit with a performance penalty if you have one stride with an offset and one without.

dlwh:
The overhead is only 4.5% as is according to a simple benchmark. I can knock it down to 2.5%, but that's about it.

Do you have an example where the manual loop over the array is much faster? It would be good to have something to measure.

PR #421, opened 2015-07-23
Fix StudentsT distribution pdf

jrj-d:
The parentheses were slightly off.

Side note:
It would be good to add a mechanism testing pdf's at a global scale in the `distributions` package.
Similarly to what is done for cdf's, we could compare the `pdf` method to the `draw` method.

About Content Square:
We are a French company aiming at improving the UX on our customers' websites. We provide insights and recommendations based on the browsing behaviors of the users. We use Scala and Spark for our data pipeline and plan to rely on Breeze and MLLib for our analysis tools. We'd be happy to contribute to Breeze on a regular basis!

http://www.contentsquare.com/en/


dlwh:
whoops.

Thanks! It's great to have more contributors!

It'd be great to have a general pdf test. Are you guys volunteering? :-)

jrj-d:
We're on it! It shouldn't take long and could help detect potential bugs in other distributions.

Thanks for your warm welcome!

PR #424, opened 2015-07-27
Unit test for pdf's

jrj-d:
A mechanism to test univariate pdf's. It is inspired by what is done for cdf's (`HasCdfTestBase`).

Along the way, several distributions were corrected:
StudentsT
Laplace
Gumbel
Pareto
Wald
Rayleigh


dlwh:
awesome thanks!

PR #429, opened 2015-08-10
Bugfix: Randomly generated long numbers are now properly constrainted.

bashimao:
Just a Bugfix for my previously committed code that makes the Long value generation behave like the RNG for Int.


bashimao:
One moment, keep this pull open. There is another bugfix coming in (one would not think how many things can go wrong with such an easy piece of code). In case you have never noticed:
<pre>
scala> Math.abs(Long.MinValue)
res4: Long = -9223372036854775808
</pre>

bashimao:
Grr.. It looks as if I have to make a new pull request...

PR #430, opened 2015-08-10
U64 to big int

ktakagaki:
@dlwh 
I think I made a mistake in the API for reading binary, vis-a-vis reading UInt64.

Previously, I was reading to Long (with an error thrown if the UInt64 value was higher than `Long.MaxValue`). I should have probably read UInt64 to `BigInt`.

I am filing a correction of this problem, but it will break the API if anyone is already using read/writeUInt64 and expects Long values... What do you think?

### DIFF
--- a/math/src/main/scala/breeze/io/RandomAccessFile.scala
+++ b/math/src/main/scala/breeze/io/RandomAccessFile.scala
@@ -612,7 +612,7 @@ class RandomAccessFile(file: File, arg0: String = "r")(implicit converter: ByteC
     * Will throw error if value < 0.
     */
   @throws(classOf[IOException])
-  final def writeUInt64(v: Long): Unit = {
+  final def writeUInt64(v: BigInt): Unit = {
     rafObj.write(converter.uInt64ToBytes(v))
   }
 
@@ -620,7 +620,7 @@ class RandomAccessFile(file: File, arg0: String = "r")(implicit converter: ByteC
     * Will throw error if value < 0.
     */
   @throws(classOf[IOException])
-  final def writeUInt64(v: Array[Long]): Unit = {
+  final def writeUInt64(v: Array[BigInt]): Unit = {
     rafObj.write( v.flatMap(converter.uInt64ToBytes(_)) )
   }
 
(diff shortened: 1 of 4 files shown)


bashimao:
Isn't that terribly slow? BigInt is a ref type. Isn't it?

dlwh:
Would be better and simpler to use spire.math.ULong

-- David

ktakagaki:
@dlwh 
That's brilliant, that's what I would have done with read/writeUInt64 in the first place if I had known of that! How about this updated PR? 
Also, I guess I should send out an email to the mailing list, since this is an API change?

@bashimao 
This should address your slowness concern and still let UInt64 work like it should

P.S. I played around a bit with microbenchmarks, I'm pretty surprised at how efficient spire.ULong is!

math/src/main/scala/breeze/io/RandomAccessFile.scala
```diff
@@ -876,6 +878,10 @@ class RandomAccessFile(file: File, arg0: String = "r")(implicit converter: ByteC
   */
 abstract class ByteConverter {
 
+  final val ZERO = BigInt("0")
```
dlwh: git rid of these?

math/src/main/scala/breeze/io/RandomAccessFile.scala
```diff
@@ -1128,18 +1140,20 @@ object ByteConverterLittleEndian extends ByteConverter  {
     tempret
   }
 
-  def uInt64ToBytes(value: Long): Array[Byte] = {
-    require(value >= 0, "Value " + value + " is out of range of 4-byte unsigned array.")
+  def uInt64ToBytes(value: ULong): Array[Byte] = {
+    require(value >= ZERO, s"Value $value is out of range of 4-byte unsigned array.")
```
dlwh: no longer holds?

math/src/main/scala/breeze/io/RandomAccessFile.scala
```diff
@@ -1128,18 +1140,20 @@ object ByteConverterLittleEndian extends ByteConverter  {
...
+    require(value <= uInt64Max, s"Value $value is out of range of 4-byte unsigned array.")
```
dlwh: also you mean 8 byte, anyhow?

dlwh:
re: notifying, sure, sounds good!

math/src/main/scala/breeze/io/RandomAccessFile.scala
ktakagaki: Done

math/src/main/scala/breeze/io/RandomAccessFile.scala
ktakagaki: Right, no longer necessary, taken out.

math/src/main/scala/breeze/io/RandomAccessFile.scala
ktakagaki: yes... :<

math/src/main/scala/breeze/io/RandomAccessFile.scala
```diff
@@ -28,8 +30,8 @@ import java.io.{File, DataInput, DataOutput, Closeable, IOException}
   * </table>
   *
   * *note: given that the JVM/Scala does not have a UInt64 type, nor a Int128 to promote to, UInt64s are dealt with in two
-  * ways... as a truncated Int64 (which will only allow half of the range of UInt64 to be actually used, but which
-  * is compatible with + - * / operations), or as a shifted Int64, where UInt64 are subtracted and shifted down to cover
+  * ways... as a BigInt (which is the Scala wrapper for the java BigInteger)
```
dlwh: ULong

ktakagaki: Fixed, sorry

dlwh:
thanks!

PR #431, opened 2015-08-11
Bugfix: Replaced Math.abs, because it does not properly map Long.MinV…

bashimao:
…alue into a positive range.

Well, I think this squashes the last bug. Why is always the easy things that make the most trouble?
Also see here: https://github.com/scalanlp/breeze/pull/429


dlwh:
thanks for sticking with it!

PR #439, opened 2015-08-31
warning free macros

calippo:
Removed annoying warnings when compiling macros.


dlwh:
Thanks @calippo for doing this!

unfortunately breeze needs to support 2.10 (as you can see in the build failure), which doesn't have the whitebox/blackbox distinction. So I either need to have two versions of the code, or accept some warnings. I decided to accept the warnings. Maybe there's something else that can be done?

PR #442, opened 2015-09-07
Bloomfilter does not work with objects with hash == Int.MinValue 

wakoan:
Bloom Filter does not work for the next test

```scala
val bloomFilter = BloomFilter.optimallySized[Int](100, 0.03)
val value = Int.MinValue
bloomFilter += value
assert(bloomFilter.contains(value))
```

### DIFF
--- a/math/src/main/scala/breeze/util/BloomFilter.scala
+++ b/math/src/main/scala/breeze/util/BloomFilter.scala
@@ -38,9 +38,9 @@ class BloomFilter[@specialized(Int, Long) T](val numBuckets: Int, val numHashFun
 
     for {
       i <- 0 to numHashFunctions
-      h = hash1 + i * hash2
-      nextHash =  if (h < 0) ~h else h
     } yield {
+      val h = hash1 + i * hash2
+      val nextHash = if (h < 0) ~h else h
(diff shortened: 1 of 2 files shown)


math/src/main/scala/breeze/util/BloomFilter.scala
```diff
@@ -32,15 +32,21 @@ class BloomFilter[@specialized(Int, Long) T](val numBuckets: Int, val numHashFun
...
-      val hash = computeHash(h, hi)
-      if(!bits.get(hash % numBuckets)) return false
-      h+= 1
+  def activeBuckets(key: T) = {
+    val hash1 = key.##
+    val hash2 = math.abs(MurmurHash3.mixLast(0, hash1))
+
+    for {
+      i <- 0 to numHashFunctions
+      h = hash1 + i * hash2
```
dlwh: in general I prefer assignments that are the end of the for comprehension to just be inside the body. creates fewer closures

dlwh:
thanks a lot! I had one little nitpick.

wakoan:
Fixed

dlwh:
thanks!

PR #462, opened 2015-10-27
replaces for loops by spire's cfor in linalg functions

nikdon:
This PR replaces for loops by spire's cfor for ```breeze.linalg``` functions and therefore increases performance of them.


dlwh:
thanks for doing this! any chance I could get you to use cforRange where possible? It's much clearer most of the time.

dlwh:
(fwiw, i bet most of these wouldn't be noticeable performance wise since they're not the inner loop)

dlwh:
note there's also cforRange2

nikdon:
Hi, David. I've replaced cfor by cforRanges where it was possible. Gains are not significant, but.. :) After CI tests I'm going to rebase it if you are ok with these proposals for PR

dlwh:
looks good.

dlwh:
awesome, thanks!

Issue #464, opened 2015-11-06
SliceMatrix does not support matrix operations


tillrohrmann:
`SliceMatrix` does not support matrix operations such as +, /, *, etc. The following code snippet does not work.

```
val idxs = Seq(0,1)
val x = DenseMatrix.rand(3,3)
val y = DenseMatrix.rand(3,3)
x(idxs,idxs)+= y(idxs, idxs) // cant find implicit conversion for += here
```

The reason is that the compiler cannot find an instance of `OpAdd.InPlace2[Matrix[Double], SliceMatrix[Double]]`. However, there is an instance of `OpAdd.InPlace2[Matrix[Double], Matrix[Double]]` which could be used by introducing a cast operation similar to the `Vector` cast operation. The casting would create operation for any subclasses of `Matrix[T]` and thus allowing all matrix operations on `SliceMatrix`

PR #465, opened 2015-11-06
Add cast for matrix operations

tillrohrmann:
Relates #464.

Breeze's matrix operations don't allow to be applied to subtypes of matrices if they are not particularly instantiated for them. Therefore this commit adds a cast operation which allows to apply operations defined for a base class to any subtype.


dlwh:
thanks!

Issue #478, opened 2016-01-04
Feature request: ODE Solver


jaketimothy:
Is this something that fits in to breeze?  It's certainly in matlab and scipy.  If so, I'd love to have a solver that works with DenseVector.  

I'd be happy to start implementing a Runge-Kutta to begin with.

So far, I've found starting points [here](http://www.scalaformachinelearning.com/2013/06/generic-runge-kutta-functional-solver.html) and in [scalation.dynamics](https://github.com/scalation/scalation/tree/master/src/main/scala/scalation/dynamics).

dlwh:
Definitely, I'd be happy to have one!

I'd probably take any of the numerical routines from that library

-- David

PR #481, opened 2016-01-17
Basic Runge-Kutta ODE solver

jaketimothy:
This is an initial implementation of an ODE solver, following issue #478.

For more ideas, see [scipy's odeint](http://docs.scipy.org/doc/scipy-0.16.0/reference/generated/scipy.integrate.odeint.html) and [matlab's odeset](http://www.mathworks.com/help/matlab/ref/odeset.html).


dlwh:
Cool, thanks!

Could you add just a bit more documentation/links? Mostly I didn't understand the array of t's until I read the matlab documentation you linked. (I don't really know ODEs.) Basically I just want a sentence or two that explains what each of the parameters does.

Also, it's not clear to me why you sometimes use Array[Double] and sometimes use DenseVector[Double].

jaketimothy:
Oh, of course!

The use of Array is something I would appreciate some input on.  DenseVector[Double] is used to hold any state vector.  The derivatives function f and the initial value y0 use DenseVector[Double].  This way, users can leverage breeze.linalg when creating the derivatives function f.  Currently, the solver uses Array as the larger container, mapping t: Array[Double] to the result: Array[DenseVector[Double]].  Thoughts?  

Now, for the error tolerance input values, I chose Array[Double] to use the convenient Array.empty and Array.isEmpty.  I would like to make them DenseVector[Double], but I'm not sure how best to implement optional DenseVector parameters.

dlwh:
ok cool! Could you add something along those lines to the comments?

I think it makes sense to do arrays like you said, in that case.

dlwh:
thanks!

Issue #126, opened 2014-01-21
NumPy parity: statistics


dlwh:
http://docs.scipy.org/doc/numpy/reference/routines.statistics.html
- [x] ptp (i'd like a better name)
- [ ] nan\* variants of the above
- [x] corrcoef
- [ ] correlate
- [x] cov
- [x] histogram
- [ ] histogram2d
- [x] bincount
- [x] digitize

dlwh:
eh, good enough

PR #487, opened 2016-01-29
Add digitize function

stucchio:
Add digitize function, equivalent to Numpy's digitize(x, bins, right=True). 

https://docs.scipy.org/doc/numpy-1.10.0/reference/generated/numpy.digitize.html

Partial resolution to #126


dlwh:
thanks! good to see you here again!

PR #491, opened 2016-01-30
Somewhat general purpose quasi-monte carlo integration

stucchio:
Here are some quasi-monte carlo codes. I've been using them for a while, been meaning to clean them up and submit to breeze for some time. 

I'm curious to hear thoughts on the structure of the code. 

Notable architectural points to think about include the way I'm loading a list of primes/etc into memory from the classpath, and the (potentially unsafe, but minimizing object creation) way I'm re-using the array of QMC output variables.

Overall this code is useful (read: makes me money in the stock market) but I'm not sure how breeze-ish it is. Maybe we should build some wrappers to make it more breezy?


math/src/main/scala/breeze/integrate/quasimontecarlo/TransformedQuasiMonteCarloGenerator.scala
```diff
@@ -0,0 +1,181 @@
...
+      val r = b+(c*y)-x
+      (r + 2.5040774 - 4.5*z >= 0) || (r >= math.log(z))
+    }
+
+    def compute(rvs: Array[Double], position: Int): Double = theta*x
+
+    def copy = GammaQuasiRandomVariableSpecAlphaGeq1(alpha, theta)
+  }
+
+  class TransformedQuasiMonteCarloGenerator(val inVariables: List[QuasiRandomVariableSpec]) extends QuasiMonteCarloGenerator {
```
dlwh: any reason to not just take a Seq or IndexedSeq on the way in?

dlwh: (i basically hate List)

math/src/main/scala/breeze/integrate/quasimontecarlo/TransformedQuasiMonteCarloGenerator.scala
```diff
@@ -0,0 +1,181 @@
...
+  trait RejectionQuasiRandomVariableSpec extends QuasiRandomVariableSpec {
+    /*
+     * The function accept will be called. If true, then the function compute will be called.
+     * This order will always be followed, so the implementer can cache values computed during
+     * accept.
+     *
+     * Of course, this does require that this trait not be used concurrently, i.e. copies of the
+     * spec always need to be created.
+     *
+     * Unfortunately in practice, a more scala-ish style of return an Option[Double] is way too slow.
```
dlwh: not saying we should do this now, but I wonder if we can get away with a custom AnyVal class that treats NaN as None. But yes, Option[Double] is unacceptable almost always

math/src/main/scala/breeze/integrate/quasimontecarlo/TransformedQuasiMonteCarloGenerator.scala
```diff
@@ -0,0 +1,181 @@
+package breeze.integrate.quasimontecarlo
+
+/*
+ Copyright 2009 Chris Stucchio
```
dlwh: is that date actually right?

math/src/main/scala/breeze/integrate/quasimontecarlo/Halton.scala
```diff
@@ -0,0 +1,109 @@
...
+   */
+
+  private val bases = java.util.Arrays.copyOfRange(Halton.PRIMES, 0, dimension)
+
+  private var count: Long = 0
+  private val counters: Array[Vector[Int]] = List.fill(dimension)({ new java.util.Vector[Int]() }).toArray
+  val permutations: Array[Array[Long]] =
+    (0 to dimension).map(i => {
+      val vv = new Array[Long](Halton.PRIMES(i))
+      cfor(0)(j => j < Halton.PRIMES(i), j => j+1)(j => {
```
dlwh: i'm generally a fan of cforRange these days

math/src/main/scala/breeze/integrate/quasimontecarlo/Halton.scala
```diff
@@ -0,0 +1,109 @@
...
+   *
+   * This generates a sequence of *uniform* quasi-random variables, i.e. variables that are
+   * can be used to approximate \int f(x) dU(x) where dU(x) is the uniform distribution on
+   * [0,1].
+   */
+
+  private val bases = java.util.Arrays.copyOfRange(Halton.PRIMES, 0, dimension)
+
+  private var count: Long = 0
+  private val counters: Array[Vector[Int]] = List.fill(dimension)({ new java.util.Vector[Int]() }).toArray
```
dlwh: since you're performance sensitive, are you not worried about the boxing overhead here?

dlwh:
Thanks!

I agree it's not the breeziest thing ever, but in general the entire distributions package needs some work. So, I think it's basically fine, modulo a few comments.

I actually implemented a Halton sequence thing like 7 years ago: https://github.com/scalanlp/breeze/blob/f4c326f9f219859156a8c55bc667369813fa4b52/math/src/main/scala/breeze/stats/random/HaltonSequence.scala


Maybe it should just be deleted in favor of yours?

math/src/main/scala/breeze/integrate/quasimontecarlo/TransformedQuasiMonteCarloGenerator.scala
stucchio: If we let it be a Seq, we get this:

```
[error] /home/stucchio/src/breeze/math/src/main/scala/breeze/integrate/quasimontecarlo/TransformedQuasiMonteCarloGenerator.scala:137: double definition:
[error] constructor TransformedQuasiMonteCarloGenerator: (inVariables: Seq[ProvidesTransformedQuasiMonteCarlo.this.QuasiRandomVariableSpec])ProvidesTransformedQuasiMonteCarlo.this.TransformedQuasiMonteCarloGenerator at line 136 and
[error] constructor TransformedQuasiMonteCarloGenerator: (inVariables: ProvidesTransformedQuasiMonteCarlo.this.QuasiRandomVariableSpec*)ProvidesTransformedQuasiMonteCarlo.this.TransformedQuasiMonteCarloGenerator at line 137
[error] have same type after erasure: (arg$outer: breeze.integrate.quasimontecarlo.ProvidesTransformedQuasiMonteCarlo, inVariables: Seq)ProvidesTransformedQuasiMonteCarlo.this.TransformedQuasiMonteCarloGenerator
[error]     def this(inVariables: QuasiRandomVariableSpec*) = this(inVariables.toSeq)
[error]         ^
[error] one error found
```

Alternately we can just eliminate the list, and just take a variable number of constructor args.

math/src/main/scala/breeze/integrate/quasimontecarlo/Halton.scala
stucchio: I am a bit. But if you look at line 91, you see that unfortunately we need a vector rather than an array since these grow occasionally. Some sort of unboxed vector type would be ideal.

dlwh: well, you could try using either HashVector or SparseVector with size = MaxInt? or i dunno. we can just do vector until you're unhappy with performance.

And why Vector instead of ArrayList/ArrayBuffer?

stucchio:
To improve performance I replaced the java.lang.Vector with an internal (and fairly unsafe) unboxed, int-only vector.

dlwh:
thanks!

PR #493, opened 2016-02-04
adds ability to set custom seed via:  RandBasis.withSeed(...)

### LINKED ISSUE #492: Setting a random seed
Hello!

I have a feature request, and would be happy to put up a PR if I receive positive feedback for it.

Currently, as far as I understand, it is a bit complicated to set a specific random seed for a probability distribution.  For instance:

```
import breeze.stats.distributions.{Uniform,RandBasis,ThreadLocalRandomGenerator}
import org.apache.commons.math3.random.{MersenneTwister}

def setSeed(seed: Int = 0) = {
      new RandBasis(new ThreadLocalRandomGenerator(new MersenneTwister(seed)))
}

val dist = new Uniform(0, 1)(setSeed())
```

I know that RandBasis.mt0 exists, but the actual seed value is an AtomicInteger, and not a configurable value.

```
val dist = new Uniform(0, 1)(RandBasis.mt0)
```
I would like to make it easier to configure the seed for Breeze distributions, and I propose defining the `setSeed` function above as a method on the RandBasis object, so we'd have `RandBasis.setSeed(seed: Int)` as an option in addition to `RandBasis.mt0`.  

I have 3 questions:

1. Is this a good idea?
2. Are there better ways or other ways to set a random seed?
3. Does it make sense for me to open a PR to add the RandBasis.setSeed method?

Thank you!

adgaudio:
closes #492 

adds RandBasis.withSeed(...) to enable functionality like:

```
import breeze.stats.distributions.{Uniform, RandBasis}

val uniform1 = new Uniform(0, 1)(RandBasis.withSeed(0))
val uniform2 = new Uniform(0, 1)(RandBasis.withSeed(0))

assert { uniform1.sample(100) == uniform2.sample(100) }
```

@dlwh you mentioned in #492 that _The atomic integer is important because with your code every thread gets its own random generator with the same seed._  I wanted to be able to have all threads use the same seed.  If users care to define a starting seed and also to ensure threads have different seeds, I think they can can wrap calls to withSeed like this `withSeed(atomicInt.getAndIncrement())`.  Can you please let me know if I misinterpreted anything, and also let me know if should change the code?

Thank you!

### DIFF
--- a/math/src/test/scala/breeze/stats/distributions/RandTest.scala
+++ b/math/src/test/scala/breeze/stats/distributions/RandTest.scala
@@ -31,6 +31,23 @@ class RandTest extends FunSuite {
     assert( RandBasis.mt0.randLong.sample(10000).forall(_ >= 0) )
   }
 
+  test("RandBasis.withSeed ensures distinct seeds in different threads") {
+    implicit val basis: RandBasis = RandBasis.withSeed(2)
+    var t2 = new Gaussian(0, 1).sample(10)
+    var t3 = new Gaussian(0, 1).sample(10)
+
+    assert { t2 != t3 }  // sanity check
+
+    val threads = for (i <- 1 to 2) yield new Thread {
+      override def run() { t2 = new Gaussian(0, 1).sample(10) }
+    }
+    threads map(_.start)
+    threads map(_.join)
+
+    // ensure that both threads use different seeds
+    assert { t2 != t3 }
+  }
+
   test("RandBasis.withSeed lets users specify a random seed") {
     {
       val a = RandBasis.withSeed(0).randInt.sample(100)
--- a/math/src/test/scala/breeze/stats/distributions/RandTest.scala
+++ b/math/src/test/scala/breeze/stats/distributions/RandTest.scala
@@ -27,8 +27,22 @@ import org.scalatest.FunSuite
  **/
 class RandTest extends FunSuite {
   test("randInt is always non-negative") {
-    RandBasis.mt0.randInt.sample(10000).forall(_ >= 0)
-    RandBasis.mt0.randLong.sample(10000).forall(_ >= 0)
+    assert( RandBasis.mt0.randInt.sample(10000).forall(_ >= 0) )
+    assert( RandBasis.mt0.randLong.sample(10000).forall(_ >= 0) )
+  }
+
+  test("RandBasis.withSeed lets users specify a random seed") {
+    {
+      val a = RandBasis.withSeed(0).randInt.sample(100)
+      val b = RandBasis.withSeed(0).randInt.sample(100)
+      assert(a == b)
+    }
+
+    {
+      val a = RandBasis.withSeed(0).randInt.sample(100)
+      val b = RandBasis.withSeed(1).randInt.sample(100)
+      assert(a != b)
+    }
   }
 
 }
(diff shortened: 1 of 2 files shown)


dlwh:
so, the way i anticipate people using these (typically) is not like you're doing but by doing

```
implicit val basis = RandBasis.seeded(7)

new Uniform(0, 1).sample(10000)
```

The fact that we're using ThreadLocalGenerator implies that we're at least thinking about the multithreaded case. Your use case will be the same with atomicinteger, and it won't be broken in the multithreaded case.

adgaudio:
Hey @dlwh ,

Unfortunately, I don't understand what you're asking me to change.  I'm new to Scala (and Java), so I need a bit more hand-holding and instruction than you seasoned veterans do! ;)  Could you please show me what you were thinking?

dlwh:
sorry for the slow response:

so in Breeze, the RandBasis is passed around as an implicit argument:

```scala
implicit val basis: RandBasis = RandBasis.withSeed(0)

val myGenerator = new Gaussian(0, 1) // (basis) is implicitly passed in rather than the default argument.
```

(The scoping rules for implicits are somewhat complex, but basically anything in lexical scope is available.)

Because they can be used this way, it's possible for the same basis to end up on two different threads. With your implementation, both threads will be given their own generator with the same initial seed, meaning they will generate precisely the same sequence of random variables. This is almost certainly bad.

If, instead, we use the atomic int thing, each thread is seeded with a unique random seed. Because the first seed is 0, the first thread that uses it will be initialized to 0, as in the current implementation.

It's possible we should figure out something that's more stable than that (since there's a data race for who gets which seed), maybe using the name of the thread(?!?), but at least it doesn't produce a catastrophic failure like generating the same variates on every thread.

So, concretely, I just want the body of the method to be this:

```scala
val int = new AtomicInteger(seed)
new RandBasis(new ThreadLocalRandomGenerator(new MersenneTwister(int.getAndIncrement())))
```

adgaudio:
Ok, thank you for your explanation!  I didn't understand that `new MersenneT...` was lazily evaluated by ThreadLocalRandomGenerator.  As far as I understand, the lazy evaluation makes it possible to do things like share the AtomicInteger, `int`, between threads.

I added your change and also a test for this.  The test calls `new Gaussian(0, 1).sample(10)` inside different threads - let me know if you see any other changes you'd like me to make.

math/src/test/scala/breeze/stats/distributions/RandTest.scala
```diff
@@ -27,8 +27,42 @@ import org.scalatest.FunSuite
...
+    assert { t2 != t3 }  // sanity check
+
+    val tt2 = new Thread {
+      override def run() { t2 = new Gaussian(0, 1).sample(10) }
+    }
+    val tt3 = new Thread {
+      override def run() { t3 = new Gaussian(0, 1).sample(10) }
+    }
+    tt2.start()
+    tt3.start()
```
dlwh: you should probably ensure they both finished...

dlwh:
thanks! i think just the one change (maybe a countdown latch?)

dlwh:
sorry i forgot to mention the laziness, that is an important part of it

math/src/test/scala/breeze/stats/distributions/RandTest.scala
adgaudio: woops - good catch.  Instead of countdownlatch, I ended up using map(_.start) and map(_.join)

dlwh:
Thanks!

merging just so it's done and because it's a test, but in general foreach is preferred to map if you're not going to use the return values

adgaudio:
Thanks @dlwh!  I'm looking forward to further contributions to Breeze - it's a really cool library!

dlwh:
great!

PR #495, opened 2016-02-07
Added sparse bincount

stucchio:
For collaborative filtering and similar applications, this is probably the more useful version. 

Please offer suggestions on the internals, I'm sure we can avoid boxing and counters.

### DIFF
--- a/math/src/main/scala/breeze/stats/DescriptiveStats.scala
+++ b/math/src/main/scala/breeze/stats/DescriptiveStats.scala
@@ -399,7 +399,6 @@ trait DescriptiveStats {
           result
         }
       }
-  }
 
   /**
     * A [[breeze.generic.UFunc]] for counting bins.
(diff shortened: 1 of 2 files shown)


dlwh:
what about either using VectorBuilder or a HashVector with length == MaxInt ?

math/src/main/scala/breeze/stats/DescriptiveStats.scala
```diff
@@ -400,6 +400,52 @@ trait DescriptiveStats {
...
+  /**
+    * A [[breeze.generic.UFunc]] for counting bins.
+    *
+    * This differs from bincount in that the result it returns is a SparseVector.
+    * The internal implementation of this could probably be significantly sped
+    * up by avoiding the use of counter. Then again, in sparse situations it's
+    * still a lot faster than using bincount.
+    *
+    */
+  object bincountSparse extends UFunc {
```
dlwh: the breeze-y thing to do, at least so far, is to make a "sparse" object within bincount. I'm definitely not tied to that, but that's how things have been so far.

stucchio:
Updated as per your suggestions.

PR #498, opened 2016-02-19
make string2file easy

DonBeo:
Hi,
I have included a `toFile` method for the string class. This should make easier to use `csvread`. 

Let me know if this is fine


dlwh:
Thanks!

can you trim the test resource a bit? unnecessarily large for what it's testing.

DonBeo:
I think it could be useful to have a standard dataset even for future test. If you prefer I can use something smaller

dlwh:
ok sure. thanks!

PR #509, opened 2016-03-14
bump shapeless 2.0.0 -> 2.2.5

declerambaul:
Bumping the shapeless version after running into versioning issues `java.lang.NoSuchMethodError: shapeless.Witness$.mkWitness(Ljava/lang/Object;)Lshapeless/Witness;` with another dependency. 

Tests pass after bumping.


dlwh:
thanks! I guess i should fix that test...

Issue #441, opened 2015-09-03
simpler syntax for csvread


danielkorzekwa:
csvread("src/test/resources/cogp/cogp_no_missing_points.csv")

in addition to existing one:
csvread(new File("src/test/resources/cogp/cogp_no_missing_points.csv"))

DonBeo:
I'm relatively new to scala but I would like to try to solve this issue. 

I think a possible solution would be something similar to the accepted answer to this question: http://stackoverflow.com/questions/3508077/how-to-define-type-disjunction-union-types 

the `file` variable should then be `string` or `java.io.File`.    
What do you think?

dlwh:
I honestly don't like using Strings as arguments to methods like that, because it's not clear if the string is going to be treated as a file, or parsed directly.

I would be onboard with an implicit that added a method like .toFile and/or file"..." for string literals:

`csvread(file"src/test/resources/cogp/cogp_no_missing_points.csv")` doesn't seem so onerous?

DonBeo:
I think this solution is better. I discovered that it is quiet complex to override a function with default values.  If this is fine for you I'll implement this change in the following days.

dlwh:
sounds great!

danielkorzekwa:
From the user point of view it would be much better to call
```csvread('somefile.csv')``` instead of ```csvread(file'somefile.csv') ```

Sounds like a little difference but when you call it many many times then it makes the difference for the user. Currently, to read csv file in both R and Scala:

R:
```
 read.csv('somefile.csv')
```

Scala:
```
import breeze.linalg._
import java.io.File
csvread(new File('somefile.csv'),skiplines=1)
```

dlwh:
I dunno, I'm inclined to stick to my guns on this one. I don't like the ambiguity between parsing a string and reading a file at a path. We have strong static types for a reason. (Maybe an additional "terse mode" import to allow what you want?)

DonBeo:
I agree with David.  In addition is quiet complex to override a function in order to make it able to accept 2 different types. 
An alternative would be to define a new function: `readlocalcsv` that takes as input a string.

frgomes:
IMHO, ``csvread(reader: java.io.Reader,...`` is preferable, since ``csvread`` could participate in a pipeline. A similar concept would be ``csvwrite(writer: java.io.Writer,...``, which can offload data to the next participant of a pipeline.

[...]

PR #516, opened 2016-03-30
#441: Implemented simple macro for simplification of file creation

kelebra:
Implemented a simple macro as suggested by @dlwh
So now we can call it like this:
`csvread(file"src/test/resources/cogp/cogp_no_missing_points.csv")`


macros/src/main/scala/breeze/macros/util.scala
```diff
@@ -0,0 +1,12 @@
+package breeze.macros
+
+import java.io.File
+
+object util {
+
+  implicit class FileMacro(val sc: StringContext) extends AnyVal {
+
+    def file(args: Any*): File = new File(sc.s(args: _*))
```
dlwh: this is great! I don't think this isn't really a macro though right?

Could we maybe just put this in, say, a new breeze.util package object?

kelebra: Totally agree, will commit today. Do you want me to squash commits afterwards?

kelebra:
Please let me know if anything else need to be done

dlwh:
thanks!

PR #519, opened 2016-03-31
Made bernoulli class RandBasis implicit

### LINKED ISSUE #518: RandBasis argument to Bernouli should be implicit
Since all other distribution classes work this way, theres no reason to not have this?

Vaishaal:
Fixes #518


dlwh:
thanks!

Issue #521, opened 2016-04-08
Surprising operator precedence


darabos:
Perhaps it's trivial to everyone, but it caused me a lot of headache until I found the bug in a complex computation.

    val d = DenseVector[Double](0.5, 0.5)
    d :* d + 1.0

The square of something, plus one. That should be greater than 1, right? No. It's `DenseVector(0.75, 0.75)` because `+` has higher precedence than `:*`.

    d :* d :+ 1.0

Yeah, that's good. Maybe I should just always use `:+` and be safe?

    1.0 :+ d :* d

No, that's `DenseVector(0.75, 0.75)` again, because `:+` and `:*` have the same precedence, so it just goes from left to right.

----

Is there a way to fix this? I think there isn't.

Is there a way to detect this and assert/warn against it? Like disallowing mixing `+` and `:*` would already avoid a class of mistakes. But it's probably too much effort for a very incomplete solution.

Is it too late to deprecate and replace these operators? Using `*:` instead of `:*` would get the precedence right I think.

If all else fails, is there a way to document this prominently? Like maybe this should be mentioned in the name of the project, it's such a fundamental pitfall :).

dlwh:
This is known and I agree pretty severe, and I intend to fix it in Breeze 0.13 by doing a deprecation cycle.

`*:`  won't work because operators ending in `:` have right associativity in Scala. So I've landed on `*:*` etc.

I'll do this at some point in the couple of months, but it's also a pretty easy thing to do. Maybe someone wants to do a PR?

darabos:
> I'll do this at some point in the couple of months, but it's also a pretty easy thing to do. Maybe someone wants to do a PR?

Sure! I'm totally new to Breeze, but I tried my best. See #522. Thanks!

darabos:
One more thing to think about is updating the documentation when 0.13 comes out. I'd love to remember and come back to do it at that time, but I know I will forget :).

kenqcl:
I am using breeze 0.13.2. when use :> to compare two DenseVector, I also get "there was one deprecation warning (since 0.13)".

I checked the API doc: http://www.scalanlp.org/api/breeze/#breeze.linalg.DenseVector
':>' is not marked as deprecated.

What is the right operator to use to compare two DenseVector?

dlwh:
`>:>`

Basically, every operator `:@` got replaced with `@:@`, except that comparison operators like `:>=` are now `>:=`

[...]

PR #522, opened 2016-04-11
Deprecate elementwise operators starting with colon to fix operator precedence

darabos:
Fixes https://github.com/scalanlp/breeze/issues/521.

Some operators look pretty wild, like `>=:>=`. Maybe I should not do this for comparison operators? Though it's good to be consistent.

For exponentiation it makes things worse, since `^:^` has _lower_ precedence than even `:^`. But `:^` had pretty low precedence to begin with, so I think it's worth doing for consistency if you don't have a better idea.

Anyway, I'm very open to comments, as I just stared using Breeze and barely know what I'm doing. Thanks for considering this!

### DIFF
--- a/math/src/main/scala/breeze/linalg/NumericOps.scala
+++ b/math/src/main/scala/breeze/linalg/NumericOps.scala
@@ -277,8 +277,8 @@ trait NumericOps[+This] extends ImmutableNumericOps[This] {
   final def :<[TT >: This, B, That](b: B)(implicit op: OpLT.Impl2[TT, B, That]) = op(repr, b)
 
   /** Element-wise less-than-or-equal-to comparator of this and b. */
-  final def <=:<=[TT >: This, B, That](b: B)(implicit op: OpLTE.Impl2[TT, B, That]) = op(repr, b)
-  @deprecated("use <=:<= instead", "0.13")
+  final def <:=[TT >: This, B, That](b: B)(implicit op: OpLTE.Impl2[TT, B, That]) = op(repr, b)
+  @deprecated("use <:= instead", "0.13")
   final def :<=[TT >: This, B, That](b: B)(implicit op: OpLTE.Impl2[TT, B, That]) = op(repr, b)
 
   /** Element-wise greater-than comparator of this and b. */
@@ -287,8 +287,8 @@ trait NumericOps[+This] extends ImmutableNumericOps[This] {
   final def :>[TT >: This, B, That](b: B)(implicit op: OpGT.Impl2[TT, B, That]) = op(repr, b)
 
   /** Element-wise greater-than-or-equal-to comparator of this and b. */
-  final def >=:>=[TT >: This, B, That](b: B)(implicit op: OpGTE.Impl2[TT, B, That]) = op(repr, b)
-  @deprecated("use >=:>= instead", "0.13")
+  final def >:=[TT >: This, B, That](b: B)(implicit op: OpGTE.Impl2[TT, B, That]) = op(repr, b)
+  @deprecated("use >:= instead", "0.13")
   final def :>=[TT >: This, B, That](b: B)(implicit op: OpGTE.Impl2[TT, B, That]) = op(repr, b)
 
 
--- a/math/src/main/scala/breeze/linalg/NumericOps.scala
+++ b/math/src/main/scala/breeze/linalg/NumericOps.scala
@@ -29,9 +29,13 @@ trait ImmutableNumericOps[+This] extends Any {
 
   // Immutable
   /** Element-wise sum of this and b. */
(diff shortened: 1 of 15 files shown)


math/src/main/scala/breeze/optimize/linear/InteriorPoint.scala
```diff
@@ -127,7 +127,7 @@ object InteriorPoint {
     diag(mat) += 1E-20
 
     val r = DenseVector.zeros[Double](m + n + m)
-    r.slice((m+n), (m+n+m)) -= (dsaff :* dzaff - sigma/m * (s dot z) )
+    r.slice((m+n), (m+n+m)) -= (dsaff *:* dzaff - sigma/m * (s dot z) )
```
darabos: I think the semantics changes here. The original code was equivalent to `dsaff :* (dzaff - sigma/m * (s dot z))`. Since there was no parenthesis to make this explicit, I suspect it was a bug, and my change fixes it. The tests pass either way, so I don't know for sure. Can you tell?

math/src/main/scala/breeze/optimize/proximal/NonlinearMinimizer.scala
```diff
@@ -166,7 +166,7 @@ object NonlinearMinimizer {
       val (f, g) = primal.calculate(x)
       val scale = x - z + u
       val proxObj = f + 0.5 * rho * pow(norm(scale), 2)
-      val proxGrad = g + scale :* rho
+      val proxGrad = g + scale *:* rho
```
darabos: I think the semantics change here as well. The original code was equivalent to `(g + scale) :* rho`. This goes against the comment at the top of the function, so I think it was a bug, and my change fixes it. The tests pass either way, so I don't know for sure. Can you please check?

math/src/main/scala/breeze/optimize/linear/InteriorPoint.scala
dlwh: nice catch!

math/src/main/scala/breeze/optimize/proximal/NonlinearMinimizer.scala
dlwh: almost certainly this new one is right.

math/src/main/scala/breeze/linalg/NumericOps.scala
```diff
@@ -254,15 +272,23 @@ trait NumericOps[+This] extends ImmutableNumericOps[This] {
...
+  @deprecated("use <=:<= instead", "0.13")
   final def :<=[TT >: This, B, That](b: B)(implicit op: OpLTE.Impl2[TT, B, That]) = op(repr, b)
 
   /** Element-wise greater-than comparator of this and b. */
+  final def >:>[TT >: This, B, That](b: B)(implicit op: OpGT.Impl2[TT, B, That]) = op(repr, b)
+  @deprecated("use >:> instead", "0.13")
   final def :>[TT >: This, B, That](b: B)(implicit op: OpGT.Impl2[TT, B, That]) = op(repr, b)
 
   /** Element-wise greater-than-or-equal-to comparator of this and b. */
+  final def >=:>=[TT >: This, B, That](b: B)(implicit op: OpGTE.Impl2[TT, B, That]) = op(repr, b)
```
dlwh: what about `>:>=` ? or `>:=`

Either seems reasonable to me.

math/src/main/scala/breeze/linalg/NumericOps.scala
```diff
@@ -79,6 +89,8 @@ trait ImmutableNumericOps[+This] extends Any {
   }
 
   /** Element-wise exponentiation of this and b. */
+  final def ^:^[TT >: This, B, That](b: B)(implicit op: OpPow.Impl2[TT, B, That]) = op(repr, b)
+  @deprecated("use ^:^ instead", "0.13")
```
dlwh: could you maybe add a note to the beginning of the string warning strings: "This operator has confusing and often surprising precedence that leads bugs. Use ... instead"

Hopefully it's a quick find replace?

Thanks!

math/src/main/scala/breeze/linalg/NumericOps.scala
darabos: > what about `>:>=` ? or `>:=`

Great idea! I went with `>:=` and `<:=`.

math/src/main/scala/breeze/linalg/NumericOps.scala
darabos: > could you maybe add a note to the beginning of the string warning strings: "This operator has confusing and often surprising precedence that leads bugs. Use ... instead"

Good point. Done.

darabos:
Thanks for the comments!

dlwh:
thanks so much! I'll merge this one travis is happy

PR #523, opened 2016-04-11
Consistent CanSlice2 Implicits

mpetnuch:
Hi. I'm submitting a pull request to fix the following issues/inconsistencies we had while using breeze internally.

1. Create TensorBoolean Slice2 implicits to mirrors the Slice implicits available for a vector in Tensor. The rational behind adding these is the have a consistent slicing interface between vectors and matrices. If you are able to do the following with a vector:

```
val v = DenseVector.tabulate(10)(x => x)
v(v :> 5)
```

You should also be able to do something similar with a matrix
```
val m = DenseMatrix.tabulate(5, 5)((i,j) => 5 * i + j + 1)
val row2Slice = m(1, ::) :> 8
m(row2Slice.t, ::)
m(row2Slice.t, 1)
m(row2Slice.t, Seq(2, 3))
```

2. The same goes for SliceMatrix. Once you have a SpliceMatrix you really can't "slice" it further. Thus I created analogous implicits to operate on slice matrixes like you can on a regular dense matrix. You should have a consistent set of available slicing methods on Matrix objects so that you are not forced to always create a copy by converting the SliceMatrix to a DenseMatrix. You should have the choice to make the copy when operating on the dense matrix makes whatever operation you are doing faster.

Let me know if you have any issues and or comments with my commit. 

Thanks!

Michael Petnuch

### DIFF
--- a/math/src/main/scala/breeze/linalg/SliceMatrix.scala
+++ b/math/src/main/scala/breeze/linalg/SliceMatrix.scala
@@ -1,5 +1,6 @@
 package breeze.linalg
 
+import breeze.linalg.operators.OpSet
 import breeze.linalg.support.CanTraverseKeyValuePairs.KeyValuePairsVisitor
 import breeze.linalg.support.CanTraverseValues.ValuesVisitor
 import breeze.linalg.support._
@@ -11,7 +12,8 @@ import scala.reflect.ClassTag
 class SliceMatrix[@specialized(Int) K1,
                   @specialized(Int) K2,
                   @specialized(Double, Int, Float, Long) V:Semiring:ClassTag](val tensor: Tensor[(K1, K2),V],
-                                                      val slice1: IndexedSeq[K1], val slice2: IndexedSeq[K2])
+                                                                              val slice1: IndexedSeq[K1],
+                                                                              val slice2: IndexedSeq[K2])
   extends Matrix[V] with MatrixLike[V, SliceMatrix[K1, K2, V]] {
 
   def apply(i: Int, j: Int): V = tensor(slice1(i)->slice2(j))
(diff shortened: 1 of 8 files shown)


math/src/main/scala/breeze/linalg/SliceMatrix.scala
```diff
@@ -131,4 +132,75 @@ object SliceMatrix {
       }
     }
   }
+
+  // slices
+  implicit def canSliceRow[V: Semiring : ClassTag]: CanSlice2[SliceMatrix[Int, Int, V], Int, ::.type, SliceVector[(Int, Int), V]] = {
+    new CanSlice2[SliceMatrix[Int, Int, V], Int, ::.type, SliceVector[(Int, Int), V]] {
+      def apply(from: SliceMatrix[Int, Int, V], row: Int, ignored: ::.type): SliceVector[(Int, Int), V] = {
+        new SliceVector(from.tensor, from.slice2.map(col => (row, col)))
```
dlwh: i don't think this is right? shouldn't you be using slice1(row) instead?

math/src/main/scala/breeze/linalg/SliceMatrix.scala
```diff
@@ -131,4 +132,75 @@ object SliceMatrix {
...
+      }
+    }
+  }
+
+  implicit def canSliceCol[V: Semiring : ClassTag]: CanSlice2[SliceMatrix[Int, Int, V], ::.type, Int, SliceVector[(Int, Int), V]] = {
+    new CanSlice2[SliceMatrix[Int, Int, V], ::.type, Int, SliceVector[(Int, Int), V]] {
+      def apply(from: SliceMatrix[Int, Int, V], ignored: ::.type, col: Int): SliceVector[(Int, Int), V] = {
+        new SliceVector(from.tensor, from.slice1.map(row => (row, col)))
```
dlwh: similarly

dlwh:
other than the one bug and the test failure (stupid 2.10 series) looks great to me.

Let me know if you have trouble with test failure and I can try to help

math/src/main/scala/breeze/linalg/SliceVector.scala
```diff
@@ -121,3 +126,34 @@ object SliceVector {
     }
   }
 }
+
+trait SliceVectorOps {
+  // todo: all all the other ops (can this be done with some macro magic?
```
dlwh: (the answer is probably, but i haven't thought about it enough yet. If it can be done with the C preprocessor, the expand macro annotations might be able to do it)

dlwh:
looks great! there must be a way to reduce some of the code duplication, but it's not obvious and breeze is chock full of it

PR #525, opened 2016-04-16
Configure external links in ScalaDoc

Atry:
I found that the ScalaDoc does not link to Scala standard library and other dependencies. This patch fixes the problem.


Atry:
By the way, we can also add the `-implicits` flag to scaladoc, that would generate correct methods added by implicit views for tensors

dlwh:
thanks! 

I'd take the -implicits PR, but relatively few methods in breeze are added via implicit views. almost all of breeze's implicit magic is via implicit parameters

Issue #515, opened 2016-03-26
Operations with differently sized DenseVectors where `a + b` fails but `b + a` succeeds


adgaudio:
Hello!

I came across this oddity and was wondering if it's expected behavior or a bug?  Specifically, we can perform operations with DenseVectors of different sizes if the small vector comes first.

```
scala> val a = DenseVector(1, 2, 3, 4, 5)
a: breeze.linalg.DenseVector[Int] = DenseVector(1, 2, 3, 4, 5)

scala> val b = DenseVector(1, 2, 3)
b: breeze.linalg.DenseVector[Int] = DenseVector(1, 2, 3)

scala> b + a
res71: breeze.linalg.DenseVector[Int] = DenseVector(2, 4, 6)

scala> a + b
java.lang.ArrayIndexOutOfBoundsException: 3
  at breeze.linalg.operators.DenseVectorOps$$anon$142.apply(DenseVectorOps.scala:161)
  at breeze.linalg.operators.DenseVectorOps$$anon$142.apply(DenseVectorOps.scala:150)
  at breeze.linalg.NumericOps$class.$plus(NumericOps.scala:172)
  at breeze.linalg.DenseVector.$plus(DenseVector.scala:51)
  ... 42 elided

```

It does seem like there should be validation that causes `b + a` to fail, though if we fixed it I bet we'd break people's downstream code that depends on this condition.  

If you want a PR for a "fix" despite my previous comment about backwards incompatibility, I'll be happy to work on this in a week or two (and please provide a couple tips where I should include the validation!).

Thanks for reviewing this!

adgaudio:
update: this doesn't work for double:

```
scala> DenseVector(1D, 2D, 3D) + DenseVector(1D, 2D, 3D, 4D, 5D)
java.lang.IllegalArgumentException: requirement failed: Vectors must have same length: x.length == y.length (5 != 3)
  at breeze.linalg.DenseVector$canDaxpy$.apply(DenseVector.scala:589)
  at breeze.linalg.DenseVector$$anon$3.apply(DenseVector.scala:581)
  at breeze.linalg.DenseVector$$anon$3.apply(DenseVector.scala:579)
  at breeze.linalg.operators.DenseVector_GenericOps$$anon$308.apply(DenseVectorOps.scala:710)
  at breeze.linalg.operators.DenseVector_GenericOps$$anon$308.apply(DenseVectorOps.scala:707)
[...]

PR #528, opened 2016-04-20
bug fix: assert that operations on different sized vectors fail. 

adgaudio:
In issue #515 I pointed out that the following bug exists:

```
scala> val a = DenseVector(1, 2, 3, 4, 5)
a: breeze.linalg.DenseVector[Int] = DenseVector(1, 2, 3, 4, 5)

scala> val b = DenseVector(1, 2, 3)
b: breeze.linalg.DenseVector[Int] = DenseVector(1, 2, 3)

scala> b + a
res71: breeze.linalg.DenseVector[Int] = DenseVector(2, 4, 6)

scala> a + b
java.lang.ArrayIndexOutOfBoundsException: 3
  at breeze.linalg.operators.DenseVectorOps$$anon$142.apply(DenseVectorOps.scala:161)
  at breeze.linalg.operators.DenseVectorOps$$anon$142.apply(DenseVectorOps.scala:150)
  at breeze.linalg.NumericOps$class.$plus(NumericOps.scala:172)
  at breeze.linalg.DenseVector.$plus(DenseVector.scala:51)
  ... 42 elided
```

In this PR, I added some assertions to ensure that both `a + b` and `b + a` will throw errors.  I also added a couple tests.  Please let me know whether I should make further changes, and thank you!


dlwh:
thanks!

PR #529, opened 2016-04-25
Refactor array util

jovilius:
Hi @dlwh,

an attempt to remove duplications in `ArrayUtil` without affecting performances. 

Does it make sense?

### DIFF
--- a/math/src/main/scala/breeze/util/ArrayUtil.scala
+++ b/math/src/main/scala/breeze/util/ArrayUtil.scala
@@ -17,14 +17,11 @@ package breeze.util
 */
 
 import java.util.Arrays
-import breeze.linalg.diff
-import breeze.macros.expand
-import spire.std.float
-
 import scala.reflect.ClassTag
-import scala.collection.mutable
 import scala.util.hashing.MurmurHash3
+import spire.std.float
 import spire.syntax.cfor._
+import breeze.macros.expand
 
 /**
  * Array operations on generic arrays, a little faster in general, I hope.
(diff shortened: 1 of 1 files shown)


math/src/main/scala/breeze/util/ArrayUtil.scala
```diff
@@ -105,46 +97,33 @@ object ArrayUtil {
...
+        case x: Array[Double] => b.asInstanceOf[Array[Double]] -> true
+        case x: Array[Float] => b.asInstanceOf[Array[Float]] -> true
+        case _ => b -> false
+      }
+      if (floating) {
+        var ai = aoffset
+        var bi = boffset
+        var i = 0
+        while (i < alength) {
+          if (a(ai) != y(bi)) return false
```
dlwh: this boxes, unfortunately

math/src/main/scala/breeze/util/ArrayUtil.scala
```diff
@@ -159,159 +138,56 @@ object ArrayUtil {
...
+        case x: Array[Byte] => b.asInstanceOf[Array[Byte]]
+        case x: Array[Boolean] => b.asInstanceOf[Array[Boolean]]
+        case x: Array[_] => b.asInstanceOf[Array[AnyRef]]
         case _ => throw new RuntimeException("shouldn't be here!")
       }
+      var ai = aoffset
+      var bi = boffset
+      var i = 0
+      while (i < alength) {
+        if (a(ai) != y(bi)) return false
```
dlwh: boxing :(

dlwh:
unfortunately this causes boxing, and possibly even reflection. If you want to do something like this, I'd recommend using the `@expand` macro like in zeroSkippingHashCodeImpl

jovilius:
Thanks, I guessed so!

jovilius:
Thanks for the feedback.

PR #531, opened 2016-05-07
Matrix Powers

claydonkey:
I was a bit surprised that there was no implementation in breeze to take the power of a matrix.
Currently there is the mpow function in breeze.linalg.functions but it fails to deliver for both Matrices with Complex Eigen values and/or Defective Matrices
I put in the rudimentary exponentiation by squaring here and also took the liberty of using the caret symbol for exp and began working on adding to the DenseMatrixOps

I suppose the best way to explain is with an example (better example in attachment):

``` scala
scala> val M = diag(DenseVector(1.0,2.0,3.0,4.0))
M: breeze.linalg.DenseMatrix[Double] =
1.0  0.0  0.0  0.0
0.0  2.0  0.0  0.0
0.0  0.0  3.0  0.0
0.0  0.0  0.0  4.0
`
scala> M ^ 2
res0: breeze.linalg.DenseMatrix[Double] =
1.0  0.0  0.0  0.0
0.0  4.0  0.0  0.0
0.0  0.0  9.0  0.0
0.0  0.0  0.0  16.0

scala> M(4,::) := DenseVector(1.0,2.0,3.0,4.0).t
res12: breeze.linalg.Transpose[breeze.linalg.DenseVector[Double]] = Transpose(DenseVector(1.0, 2.0, 3.0, 4.0))

scala> M
res13: breeze.linalg.DenseMatrix[Double] =
1.0  0.0  0.0  0.0
0.0  2.0  0.0  0.0
0.0  0.0  3.0  0.0
1.0  2.0  3.0  4.0

scala> M ^ 2
res15: breeze.linalg.DenseMatrix[Double] =
1.0                0.0   0.0                 0.0
0.0                4.0   0.0                 0.0
0.0                0.0   9.0                 0.0
4.999999999999999  12.0  21.000000000000004  16.0



[Breeze_Matrix_Exp.txt](https://github.com/scalanlp/breeze/files/253565/Breeze_Matrix_Exp.txt)
```


dlwh:
Hi Clay, 

Thanks for this!

I can't tell what's going on in the files, but github's diff is doing something really weird (it looks like math/build.sbt is replaced with an exactly identical file??). Can you see if you can rebase/squash or something to see if you can make github's diff happier?

claydonkey:
Yeah I'm having problems with netbeans producing the wrong whitespace.
I will fix this if you wish.
I the meantime you can view as it's supposed to be by amending the github url with 'w=1' to ignore whitespaces  i.e 
https://github.com/scalanlp/breeze/pull/531/commits/51f51862497e110e3010fda7bfc9ca1c3ef62b7b?w=1

claydonkey:
OK we're back in business - classic CRLF<==>LF file format  f-up.
There are a couple of points I'd like to say about the amendments.
Firstly i wouldn't imagine this would be merged as is because the element-wise exponentiation -  :^ and ^:^ is changed to Matrix exponentiation.

I was going to put a new OpType in. Something like 
sealed trait OpPowMatrix extends OpType
object OpPowMatrix extends OpPowMatrix with UFunc
and then manage the operation presidence for a Matrix Power
@expand.sequence[Op]({_ + _},  {_ - _}, {_ * _}, {_ * _}, {_ / _}, {_ % _}, {_ pow _}, {_ ^ _} ' <== here 

Which leads to another question - What symbol is correct for a matrix exp?
I realize that Scala usually reserves the caret ^ for bit-wise ops but this is already taken care of in breeze by using ^^ 
Perhaps a Matlab ** would be better?

math/src/main/scala/breeze/linalg/functions/mpow.scala
```diff
@@ -9,14 +9,25 @@ import breeze.linalg.eig.Eig
  */
 object mpow extends UFunc {
 
+  def pow(exp: Double, value: DenseMatrix[Double]): DenseMatrix[Double] = {
```
dlwh: reorder arguments to be consistent? maybe make private?

dlwh: what are you going to do if exp is not an integer?

dlwh: also what if exp is nonpositive?

math/src/main/scala/breeze/linalg/operators/DenseMatrixOps.scala
```diff
@@ -18,7 +18,79 @@ import spire.syntax.cfor._
...
+import breeze.generic.UFunc
+import breeze.linalg.eig.Eig
+
+/*@author acampbell */
+trait DenseMatrixPowerStuff extends DenseMatrixOps
+			       with DenseMatrixPowerOps
+			       with LowPriorityDenseMatrix { this: DenseMatrix.type =>
+  // <editor-fold defaultstate="collapsed" desc=" OpPow implementations ">
+
+  def expBySq(exp: Double, mat: DenseMatrix[Double]): DenseMatrix[Double] = {
```
dlwh: same concerns

math/src/main/scala/breeze/linalg/operators/DenseMatrixOps.scala
```diff
@@ -18,7 +18,79 @@ import spire.syntax.cfor._
...
+    if (exp == 1) mat
+    else if (exp % 2 == 1) mat * expBySq(exp - 1, mat)
+    else {
+      val h = expBySq(exp / 2, mat)
+      h * h
+    }
+  }
+
+implicit object implOpPow_DMD_Deq_DMD
+extends OpPow.Impl2[DenseMatrix[Double], Double, DenseMatrix[Double]] {
```
dlwh: this implementation should be shared between this and mpow

math/src/main/scala/breeze/linalg/operators/DenseMatrixOps.scala
```diff
@@ -18,7 +18,79 @@ import spire.syntax.cfor._
...
+  def apply(_a: DenseMatrix[Double], _b: Int): DenseMatrix[Double] = {
+    implOpPow_DMD_Deq_DMD(_a,_b)
+  }
+}
+
+implicit object implOpPow_IMI_Ieq_IMI
+extends OpPow.Impl2[DenseMatrix[Int], Int, DenseMatrix[Int]] {
+  def apply(_a: DenseMatrix[Int], _b: Int): DenseMatrix[Int] = {
+    val t = DenseMatrix((0, 0, 0), (0, 0, 0), (0, 0, 0))
+    t
```
dlwh: ???

math/src/main/scala/breeze/linalg/operators/DenseMatrixOps.scala
```diff
@@ -18,7 +18,79 @@ import spire.syntax.cfor._
...
+  // <editor-fold defaultstate="collapsed" desc=" implicit implementations for BinaryRegistry ">
+
+  @expand
+  @expand.valify
+  implicit def op_DM_D[@expand.args(Int, Long, Float, Double) T]: BinaryRegistry[DenseMatrix[T], T, OpPow.type, DenseMatrix[T]] =
+
+    new BinaryRegistry[DenseMatrix[T], T, OpPow.type, DenseMatrix[T]] {
+      override def bindingMissing(a: DenseMatrix[T], b: T): DenseMatrix[T] = {
+
+        val res: DenseMatrix[T] = DenseMatrix.zeros[T](a.rows, a.rows)
```
dlwh: ???

dlwh:
Yeah, I'm not sure what operator to use. I'm loath to use ** because of the (new) ` *:* ` operator, but the world seems to have converged on it, so we should probably follow suit.

I agree with the general plan of introducing a PowMatrix optype.

The expand stuff (btw, don't use unquoted `@expand` because it emails the use with that name) doesn't set precedence; it's just the order we do it in. Scala's operator precedence is defined by the first character of the operator, usually.

dlwh:
this seems useful: http://www.maths.manchester.ac.uk/~lijing/papers/SIMAX11.pdf

math/src/main/scala/breeze/linalg/operators/DenseMatrixOps.scala
claydonkey: Sorry about this. This is just me experimenting

- Is **implOpPow_DMD_Ieq_DMD** ok (as an overload)?
- Is there any point in creating  **implOpPow_IMI_Ieq_IMI** - an integer only function? 
 1. 
Is it even possible ?
 2. 
 Does it ditch the need for floating point, which would be the only reason for it - I suppose

I was trying to understand why [T:Numeric] can't be used as a generic type to create a one size fits all function instead of why each function has to have its generic argument types declared explicitly and ultimate behaves more like classic OO overloading.

On closer inspection I understand that trying to create a function like: 
```scala
  implicit def implOpPow_DMT_B_eq_DMT[T,B](implicit op: OpPow.Impl2[DenseMatrix[T], B, DenseMatrix[T]]):
  OpPow.Impl2[DenseMatrix[T], B,  DenseMatrix[T]] =
   new OpPow.Impl2[DenseMatrix[T], B, DenseMatrix[T]] {
      override def apply(_a: DenseMatrix[T], _b: B): DenseMatrix[T] = {
	requireSquareMatrix(_a)
	val Eig(real, imag, evectors) = eig(_a)
	...
	}
      }
```
is a futile because the singleton eig requires a DenseVector[Double]

math/src/main/scala/breeze/linalg/operators/DenseMatrixOps.scala
claydonkey: Again I'm sorry about this. This is more for me than a serious addition...
I need a little help understanding BinaryRegistry...

math/src/main/scala/breeze/linalg/functions/mpow.scala
claydonkey: Consider these shortcomings as being addressed. The only reason I hadn't implemented any more sophisticated was that my needs were trivial (I just needed integer exponentiation of a a Matrix of Doubles) but since were getting serious :)

claydonkey:
Brilliant.
Thanks for the critique. It was a very down and dirty implementation.
I will endeavour to create a more complete implementation.
Also I would like to see a covariance function. So I'll crack on with that as well...

PR #537, opened 2016-05-19
Consistent CanSlice2 for Row Slices of Matrix returning Transpose

mpetnuch:
When I submitted my previous pull request I didn't have the row slices return transposed vectors. This fixes that and makes them all consistent.


dlwh:
thanks!

PR #540, opened 2016-05-23
Windowed UFuncs

mpetnuch:
I'm attempting at creating windowing support for vectors in a similar vein to broadcasting for matrices. Essentially the idea was that you may want to evaluate rolling type statistics on a vector. Instead of creating separate windowed variants of the stat UFunc, we create a new UFunc which can window the underlying type.

For example, to create a rolling mean on a vector with a window length of 5 you could do the following:

```
val vector = ...
val rollingMeans = mean(window(vector, 5)) 
```

The "windowed" UFunc is just a mapping UFunc into objects of the type "Windowed" which is very similar to the Broadcasted trait.

I've got most of this working. Still thinking about how (or if I even want to handle broadcasted columns and rows). However there are some minor issues that I'm having with the implicits. The first (and larger issues) is the implicit associated with the windowed UFunc. Everything works as expected if I explicitly pull in the values. However, if I do not, I keep on getting compile errors about a missing implicit for the windowed UFunc (see windowTest.scala for what I am talking about). 

The second issue is with the CanPan implicit. I just can't figure out why the compile can't find it.

I was staring at it pretty much all weekend and I just don't see.  Hoping that maybe another set of eyes will help make it apparent. If you have some time to look this over it would be greatly appreciated.

Thanks!


math/src/main/scala/breeze/linalg/DenseVector.scala
```diff
@@ -41,8 +46,7 @@ import scalaxy.debug._
  * The i'th element is at offset + i * stride
  *
  * @author dlwh
- *
- * @param data data array
+  * @param data data array
```
dlwh: odd whitespace change

math/src/main/scala/breeze/linalg/DenseVector.scala
```diff
@@ -119,7 +123,8 @@ class DenseVector[@spec(Double, Int, Float, Long) V](val data: Array[V],
 
   /**
    * Returns a copy of this DenseVector. stride will always be 1, offset will always be 0.
-   * @return
+    *
```
dlwh: same, and throughout

math/src/main/scala/breeze/linalg/Vector.scala
```diff
@@ -62,6 +65,10 @@ trait Vector[@spec(Int, Double, Float) V] extends VectorLike[V, Vector[V]]{
 
   def keysIterator = Iterator.range(0, size)
 
+  def slice(from: Int, end: Int, stride: Int) : Vector[V] = ???
+  /// would like to do this but we have no class tag for V
+  /// new DenseVector(from.to(end).by(stride).map(i => apply(i)).toArray[V])
```
dlwh: slices elsewhere in breeze have view semantics, so I'd be somewhat hesitant to do it this way

dlwh: why not just leave it abstract?

math/src/main/scala/breeze/linalg/Vector.scala
```diff
@@ -857,10 +870,24 @@ trait VectorConstructors[Vec[T]<:Vector[T]] {
 //     apply(values.toArray)
   }
 
+  /**
+    * Allow sequences to be converted into vectors
```
dlwh: nice

math/src/main/scala/breeze/linalg/Windowed.scala
```diff
@@ -0,0 +1,128 @@
+package breeze.linalg
+
+import breeze.generic.UFunc.{InPlaceImpl, InPlaceImpl2, UImpl, UImpl2}
+import breeze.linalg.support.{CanCollapseWindow, CanForeachValues, CanIterateWindow, CanMapValues, CanTraverseWindow, ScalarOf}
+
+import scala.language.higherKinds
```
dlwh: i think i just have language:_ enabled in compile options because I don't like the training wheels stuff, but whatever

math/src/test/scala/breeze/linalg/functions/windowTest.scala
```diff
@@ -0,0 +1,34 @@
...
+    val op = mean.reduce_Double[DenseVector[Double]]
+    val handHold = DenseVector.handholdCanWindowDenseVector[Double]
+    val canCollapseWindow = DenseVector.canCollapseWindow[Double, Double]
+
+    val windowedMeanOp = WindowedVector.windowedOp(handHold, canCollapseWindow, op)
+    val windowedMeans = windowedMeanOp.apply(window(vector, 20))
+    println(windowedMeans)
+
+    // this fails with a compiler error
+    // that it cannot find implicit for impl: mean.Impl[WindowedDenseVector[Double], DenseVector[Double]]
```
dlwh: have you looked at -XlogImplicits? also, I usually work with these kinds of errors by using implicitly to summon the various pieces until one doesn't work.

math/src/main/scala/breeze/linalg/WindowedVector.scala
```diff
@@ -0,0 +1,86 @@
...
+/**
+  * DenseVector implementation of windowed vector iterator
+  */
+private case class WindowedDenseVectorIterator[T: ClassTag : Semiring](underlying: DenseVector[T], window: Window)
+  extends WindowedVectorIterator[T, DenseVector[T], DenseVector[T]](underlying, window) {
+
+  def sliceUnderlying(start: Int, end: Int): DenseVector[T] = underlying.slice(start, end)
+
+  def padResult(chunk: DenseVector[T], length: Int, padMode: OptPadMode): DenseVector[T] = {
+    //    this doesn't work as scala complains it can't find the canPad implicit, any idea???
```
dlwh: oh, this is dumb. implDV_OptPadDim is under `@expand` so only the types listed there work. We'd need to replace the implementation with one that doesn't use `@expand`.

It might be easier to just require a CanPadRight implicit in the constructor...

dlwh:
definitely don't know what's going on with the first, I'll try to take a look this week.

math/src/main/scala/breeze/linalg/DenseVector.scala
mpetnuch: Weird, I don't know why intellij decided to reform this like that. Will revert.

math/src/main/scala/breeze/linalg/Vector.scala
mpetnuch: Looking over it again, I don't need it. For some reason after seeing the slice method in dense matrix I totally forgot the CanSlice implicits and can just replace with:

` vector(a to b) `

math/src/main/scala/breeze/linalg/WindowedVector.scala
mpetnuch: Took the easy way out and added it to the constructor and moved the implicit resolution to the `canXXXWindow` functions which need the iterator.

This is tangental to what I'm doing here but it seems the "right" thing to do would be create separate implicits depending on the capabilities of the types.
- value/wrap/reflect => require nothing
- zero => requires SemiRing
- min/max/median => requires a total order 
- mean => would only be defined for standard numeric types 

Then you would of course need to update the min/max/mediam/mean UFuncs.

mpetnuch:
Okay so it looks like I've established the implicit that is causing a problem. It is WindowedOps.windowedOp

It is defined as follows:

`
  implicit def windowedOp[Op, UnderlyingType, WindowedType, CollapsedType, ResultType]
  (implicit handhold: CanCollapseWindow.HandHold[UnderlyingType, WindowedType],
   canCollapseWindow: CanCollapseWindow[UnderlyingType, WindowedType, CollapsedType, ResultType],
   op: UImpl[Op, WindowedType, CollapsedType]) = {
    new UImpl[Op, CC[UnderlyingType, WindowedType], ResultType] {
      def apply(v: CC[UnderlyingType, WindowedType]): ResultType = {
        canCollapseWindow(v.underlying, v.window) {
          op(_)
        }
      }
    }
  }
`


For some reason the type being inferred for Collapsed type is Nothing:

`<console>:19: linalg.this.WindowedVector.windowedOp is not a valid implicit value for breeze.linalg.sum.Impl[breeze.linalg.WindowedVector[breeze.linalg.DenseVector[Double],breeze.linalg.DenseVector[Double]],Double] because:
hasMatchingSymbol reported error: type mismatch;
 found   : breeze.linalg.support.CanCollapseWindow[breeze.linalg.DenseVector[Double],breeze.linalg.DenseVector[Double],Nothing,breeze.linalg.DenseVector[Nothing]]
 required: breeze.linalg.support.CanCollapseWindow[breeze.linalg.DenseVector[Double],breeze.linalg.DenseVector[Double],Double,breeze.linalg.DenseVector[Nothing]]`


If I create a specific one for the collapsed type as Double everything works. But I don't want to do that :)

Ideas?

dlwh:
ah, right, you're missing the point of HandHold (which is totally
non-obvious and exists because Scala's implicit resolution mechanism is
kind of dumb)

I don't have a ton of time at the moment. Could you read
http://dlwh.org/scala-implicit-search/ and see if that helps?

The trick is that you have to give the implicit resolution mechanism the
type of Collapsed the very first time it's mentioned in the implicit
parameter list. If you don't, it will infer Nothing

mpetnuch:
Thanks for much for writing the blog post. After re-ordering the implicits to help out the compiler it works now :)

mpetnuch:
I'll work on supporting windowing for Sparse and Slice vectors and code up some tests.

What's your thoughts on also expanding this to Broadcasted rows/columns? I think I might stop after I get it all working for "normal" vector types and then make a second PR for the broadcasted types (as there are some issues with the BroadcastedColumn type incorrectly reporting their values as being column vectors when they are actually row vectors).

dlwh:
awesome, glad it helped!

re: broadcasting, i think that makes a lot of sense.

Issue #556, opened 2016-07-04
error in sbt command 


DonBeo:
I'm not an expert but I think there are a couple of comas ',' in the wrong position in the home page:

```
libraryDependencies  ++= Seq(
  // other dependencies here
  "org.scalanlp" %% "breeze" % "0.12",
  // native libraries are not included by default. add this if you want them (as of 0.7)
  // native libraries greatly improve performance, but increase jar sizes. 
  // It also packages various blas implementations, which have licenses that may or may not
  // be compatible with the Apache License. No GPL code, as best I know.
  "org.scalanlp" %% "breeze-natives" % "0.12",
  // the visualization library is distributed separately as well. 
  // It depends on LGPL code.
    "org.scalanlp" %% "breeze-viz" % "0.12"

    // To depend on snapshot versions, use:
      "org.scalanlp" %% "breeze" % "latest.integration",
)


```

PR #557, opened 2016-07-05
#556 - update README.md

nikdon:
Fix errors in REAMDE.md related to the project definition. In addition to that, there are several updates to easy distinguish snapshot and last stable release configs.


dlwh:
thanks!

Issue #561, opened 2016-07-14
argtopk failed when applying to DenseVector and k == length of vector


yanboliang:
```argtopk``` failed when applying to ```DenseVector``` and ```k == length of vector```. The 0.12 release introduced this bug.
```
val topicCounts = DenseVector(Array(1.29, 4.89, 0.8, 1.8, 2.5, 0.3))
val topIndices = argtopk(topicCounts, 6)
println(topIndices.toArray.mkString(","))
```
It throws exception:
```
Exception in thread "main" java.lang.IllegalArgumentException: requirement failed: Invalid position specification: 6 with coll length: 6
	at scala.Predef$.require(Predef.scala:224)
	at breeze.util.quickSelect$$anon$5.implQuickSelectSort$5(Selecting.scala:103)
	at breeze.util.quickSelect$$anon$5.apply$mcI$sp(Selecting.scala:137)
	at breeze.linalg.argtopk$$anon$5.apply(argsort.scala:40)
	at breeze.linalg.argtopk$$anon$5.apply(argsort.scala:36)
	at breeze.generic.UFunc$class.apply(UFunc.scala:53)
	at breeze.linalg.argtopk$.apply(argsort.scala:33)
```

yanboliang:
Send PR #562 to fix this bug.

yanboliang:
#562 fix this bug, so close this issue.

PR #562, opened 2016-07-14
Fix bug of argtopk applying to DenseVector and k == length of vector

yanboliang:
```argtopk``` failed when applying to ```DenseVector``` and ```k == length of vector```. See more detail about this bug at #561.
We speed up ```argtopk``` for ```DenseVector``` by ```quickSelect```, but the argument of ```quickSelect``` is the vector indices which start with 0 and end with length - 1. So k - 1 should be pass in rather than k. This PR fix this bug.

### DIFF
--- a/math/src/main/scala/breeze/linalg/functions/argsort.scala
+++ b/math/src/main/scala/breeze/linalg/functions/argsort.scala
@@ -26,7 +26,7 @@ object argsort extends UFunc with LowPriorityArgSort {
 
 
 /**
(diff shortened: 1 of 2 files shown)


yanboliang:
@dlwh Would you mind to take a look when you have time? It looks like the CI build failure is uncorrelated with this PR. Thanks!

dlwh:
thanks!

dlwh:
i do need to fix the travis build, but...

PR #564, opened 2016-07-15
fix length of BroadcastedDMRowsISeq

tek:
both broadcasted ISeq adapters use `.col`, for non-square matrices, this
truncates or throws


dlwh:
Thanks!

PR #566, opened 2016-07-22
Fix the convergence check of L-BFGS-B.

yanboliang:
If there is ```searchFailed``` during optimization by ```L-BFGS-B```, it should be terminated rather than calling in an endless loop. The convergence check of ```L-BFGS-B``` does not handle ```searchFailed``` and ```gradientConverged``` currently. Since ```L-BFGS-B``` extends from ```FirstOrderMinimizer```, so my fix is let it call ```FirstOrderMinimizer.defaultConvergenceCheck``` to handle convergence check.


yanboliang:
cc @dlwh

dlwh:
thanks!

PR #571, opened 2016-08-02
Modify shuffle.scala and add shuffleTest.scala

hl475:
The original `shuffle.scala` was supposed to implement the [Fisher–Yates shuffle](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle). Unfortunately, there was a bug that prevented the code from producing uniformly distributed permutations (whereas the purpose of the Fisher-Yates shuffle is to produce uniformly distributed permutations). The attached file, [TestShuffle.zip](https://github.com/scalanlp/breeze/files/394972/TestShuffle.zip), runs a quick test that demonstrates the problem; the test creates an array of integers from 1 to 10, shuffles the array 100,000 times, and creates a histogram to record the first entry of the integer array. If the shuffling is uniformly distributed, each number should appear roughly 100,000/10 = 10,000 times within certain ranges. 

The output from the previously committed code is:
`DenseVector(10073, 12883, 12008, 11144, 10389, 9861, 8983, 8618, 8256, 7785)`
The output from the new code is:
`DenseVector(10023, 10022, 10022, 10001, 10025, 10032, 10061, 10065, 9837, 9912)`
There are ten entries of the output array; each of them represents the number of times each of the ten numbers has appeared in the first entry of the shuffled array. If the shuffle is uniformly random, then each of these should be close to 10,000 within a few standard deviations (the standard deviation is sqrt(100,000) = 316.2). The second and third entries of the output array from the previously committed code are certainly far outside of a few standard deviations, whereas the output from the new codes produces reasonable results.

We fix the bug and add an additional feature to the committed `shuffle.scala` code: the new code can return the given Array as a shuffled copy with the corresponding shuffle index information, or return the given Array as a shuffled copy using the inverse of the given shuffle index information, reversing the shuffle. Furthermore, we add a unit test via the file `shuffleTest.scala` in the directory `breeze/math/src/test/scala/breeze/linalg/functions` to test that the shuffle is uniformly random (as above); the unit test also shuffles forward and back to test the new features as well.

### DIFF
--- a/math/src/main/scala/breeze/linalg/functions/shuffle.scala
+++ b/math/src/main/scala/breeze/linalg/functions/shuffle.scala
@@ -1,64 +1,145 @@
 package breeze.linalg
 
 import breeze.generic.UFunc
-import breeze.macros.expand
-import spire.math.Complex
 import scala.collection.generic.CanBuildFrom
 import scala.collection.mutable.ArrayBuffer
 import scala.reflect.ClassTag
 import breeze.stats.distributions.Rand
 
-/**Returns the given DenseVector, Array, or DenseMatrix as a shuffled copy. (Fisher-Yates shuffle)
+/**
+  * Return the given DenseVector, Array, or DenseMatrix as a shuffled copy
+  * by using Fisher-Yates shuffle.
+  * Additionally, can return the given Array as a shuffled copy with the
+  * corresponding shuffle index information, or return the given Array as a
+  * shuffled copy using the inverse of the given shuffle index information,
+  * reversing the shuffle.
+  *
   * @author ktakagaki
-  * @date 05/12/2014.
+  * @date 05/12/2014
+  * @author huaminli
+  * @date 08/01/2016
   */
 object shuffle extends UFunc {
 
-  implicit def implShuffle_Arr_eq_Arr[T](implicit ct: ClassTag[T]): Impl[Array[T], Array[T]] = {
+  implicit def implShuffle_Arr_eq_Arr[T](implicit ct: ClassTag[T]):
+  Impl[Array[T], Array[T]] = {
     new Impl[Array[T], Array[T]] {
-
+      /**
+        * Shuffle the given [[Array[T]]].
+        *
+        * @param arr the given array stored as [[Array[T]]]
+        * @return a shuffled array stored as [[Array[T]]]
+        */
       def apply(arr: Array[T]): Array[T] = {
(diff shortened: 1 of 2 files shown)


hl475:
@dlwh, everything works fine with the latest version of Scala (2.11.8) if I run
`sbt -Dcom.github.fommil.netlib.BLAS=com.github.fommil.netlib.F2jBLAS -Dcom.github.fommil.netlib.LAPACK=com.github.fommil.netlib.F2jLAPACK -Dcom.github.fommil.netlib.ARPACK=com.github.fommil.netlib.F2jARPACK test` in my local laptop, whereas even master breaks with Travis CI build's version if I run `sbt -Dcom.github.fommil.netlib.BLAS=com.github.fommil.netlib.F2jBLAS -Dcom.github.fommil.netlib.LAPACK=com.github.fommil.netlib.F2jLAPACK -Dcom.github.fommil.netlib.ARPACK=com.github.fommil.netlib.F2jARPACK ++2.10.3`. What could be the problem here? Why do we need to specify `++2.10.3`?

dlwh:
Thanks! I'll take a look at the PR soon.

Something broke the 2.10 build a while back and I haven't had the time or
inclination to fix

math/src/main/scala/breeze/linalg/functions/shuffle.scala
```diff
@@ -1,64 +1,145 @@
...
+  Impl3[Array[T], Array[Int], Int, Array[T]] = {
+    new Impl3[Array[T], Array[Int], Int, Array[T]] {
+      /**
+        * shuffle the given [[Array[T]]] according to the given
+        * permutation [[Array[Int]]] if [[Int]] indicator is zero;
+        * shuffle the given [[Array[T]]] according to the inverse of the given
+        * permutation [[Array[Int]]] if [[Int]] indicator is nonzero.
+        * @param arr the given array stored as [[Array[T]]]
+        * @param arrIndex the given permutation array stored as [[Array[Int]]]
+        * @param isInverse the indicator whether perform inverse shuffle
```
dlwh: why not a boolean?

dlwh:
one quick question. otherwise looks great!

hl475:
You are right! I have changed `int` from `implShuffle_Arr_Arr_Int_eq_Arr` of `shuffle.scala` to `boolean`.

dlwh:
thanks!

PR #575, opened 2016-08-19
use a less confusing name

zasdfgbnm:
If you use `eta` here as name, no one can tell what it means from scaladoc.


dlwh:
Thanks!

PR #579, opened 2016-08-25
Add rectifier function (Relu).

carschno:
Add element-wise[ rectifier function](https://en.wikipedia.org/wiki/Rectifier_(neural_networks)).


dlwh:
thanks!

PR #580, opened 2016-08-30
use local instance randInt to utilize provided generator

brendon-:
The `RandInt.choose` method did not use the instance provided `generator`, but instead used the Rand object.  This causes a call to `myRandBasis.choose` to create a `Rand` that doesn't use the instances random stream.


dlwh:
thanks!

PR #582, opened 2016-09-13
Add capability to perform basic symbolic computations

nightscape:
This PR is _not yet ready to merge_ (it lacks tests, documentation and some improvements), but we'd like to get early feedback and code reviews.
It contains a set of SymbolicFunction classes and type classes to perform derivation, simplification, vectorization and conversion into ordinary Scala functions on them.
Currently, all computations are done by the compiler itself, although one could consider the use of macros which employ some JVM-based Computational Algebra System to perform advanced simplifications.

We're happy about all forms of feedback, especially about ideas on how to perform a function simplification that doesn't stop after one pass.

### DIFF
--- a/math/src/main/scala/breeze/symbolic/CanBuildSingleArgumentFunction.scala
+++ b/math/src/main/scala/breeze/symbolic/CanBuildSingleArgumentFunction.scala
@@ -1,0 +1,16 @@
+package breeze.symbolic
+
+trait CanBuildSingleArgumentFunction[A, F[A <: SymbolicFunction[A]] <: SymbolicFunction[F[A]]] {
+  def build[I <: SymbolicFunction[I]](i: I): F[I]
+}
+
+object CanBuildSingleArgumentFunction {
+  implicit def canBuildExponential[A]: CanBuildSingleArgumentFunction[A, Exponential] = new CanBuildSingleArgumentFunction[A, Exponential] {
+      override def build[I <: SymbolicFunction[I]](i: I): Exponential[I] = new Exponential(i)
+    }
+
+  implicit def canBuildLogarithm[A]: CanBuildSingleArgumentFunction[A, Logarithm] = new CanBuildSingleArgumentFunction[A, Logarithm] {
+      override def build[I <: SymbolicFunction[I]](i: I): Logarithm[I] = new Logarithm(i)
+    }
+
+}
--- a/math/src/main/scala/breeze/symbolic/CanCreateFunction.scala
+++ b/math/src/main/scala/breeze/symbolic/CanCreateFunction.scala
@@ -1,0 +1,234 @@
+package breeze.symbolic
+
+import breeze.generic.{MappingUFunc, UFunc}
+import breeze.linalg.DenseVector
+import breeze.linalg.operators.{OpAdd, OpDiv, OpMulScalar, OpMulMatrix, OpSub}
+import breeze.linalg.support.CanCreateZerosLike
+import breeze.numerics
+import shapeless.LUBConstraint._
+import shapeless.{::, HList, HNil}
+
+/**
+  * Type-class to create an ordinary Scala function from a SymbolicFunction
+  * @tparam I The input type of the Scala function
+  * @tparam O The output type of the Scala function
+  * @tparam F The type of the SymbolicFunction from which the Scala function is created
+  */
+trait CanCreateFunction[I, O, F <: SymbolicFunction[F]] {
+  def createFunction(f: F): I => O
+}
+
+trait LowPriorityCanCreateFunction {
+  implicit def canCreateVectorizedSymbolicFunction[I, O, F <: SymbolicFunction[F]](
+    implicit canCreateFunction: CanCreateFunction[I, O, F],
+    bf: breeze.linalg.support.CanMapKeyValuePairs[DenseVector[I], Int, I, O, DenseVector[O]]
+  ): CanCreateFunction[DenseVector[I], DenseVector[O], VectorizedSymbolicFunction[F]] =
+    new CanCreateFunction[DenseVector[I], DenseVector[O], VectorizedSymbolicFunction[F]] {
+      override def createFunction(f: VectorizedSymbolicFunction[F]): (DenseVector[I]) => DenseVector[O] = {
+        val convertedFunctions = f.fs.map(canCreateFunction.createFunction).toArray
+        (i: DenseVector[I]) => i.mapPairs { case (index, value) => convertedFunctions(index).apply(value) }
+      }
+    }
+
+  implicit def canCreateEmptyProductFunction[I, O](
+    implicit constOne: CanCreateFunction[I, O, ConstOne]
+  ): CanCreateFunction[I, O, Product[HNil]] =
+    new CanCreateFunction[I, O, Product[HNil]] {
+      def createFunction(f: Product[HNil]) = constOne.createFunction(ConstOne())
+    }
+
+  implicit def canCreateProductScalarFunction[I, O1, O2, F <: SymbolicFunction[F], L <: HList : <<:[SymbolicFunction[_]]#λ](
+    implicit ccf1: CanCreateFunction[I, O1, F],
+    ccf2: CanCreateFunction[I, O2, Product[L]],
+    multiply: OpMulMatrix.Impl2[O1, O2, O1]
+  ) = new CanCreateFunction[I, O1, Product[F :: L]] {
+    def createFunction(f: Product[F :: L]) = {
+      val f1 = ccf1.createFunction(f.fns.head)
+      val f2 = ccf2.createFunction(Product(f.fns.tail))
+                                  (i: I) =>
+      multiply(f1(i), f2(i))
+    }
+  }
+
+  implicit def canCreateEmptySumFunction[I, O](
+    implicit constZero: CanCreateFunction[I, O, ConstZero]
+  ): CanCreateFunction[I, O, Sum[HNil]] =
+    new CanCreateFunction[I, O, Sum[HNil]] {
+      def createFunction(f: Sum[HNil]) = constZero.createFunction(ConstZero())
+    }
+
+  implicit def canCreateSumFunction[I, O, S1 <: SymbolicFunction[S1], S2 <: HList : <<:[SymbolicFunction[_]]#λ](
+    implicit ccf1: CanCreateFunction[I, O, S1],
+    ccf2: CanCreateFunction[I, O, Sum[S2]],
+    add: OpAdd.Impl2[O, O, O]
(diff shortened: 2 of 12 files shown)


math/src/main/scala/breeze/symbolic/CanBuildSingleArgumentFunction.scala
```diff
@@ -0,0 +1,18 @@
+package breeze.symbolic
+
+trait CanBuildSingleArgumentFunction[F[_ <: SymbolicFunction] <: SymbolicFunction] {
+  def build[I <: SymbolicFunction](i: I): F[I]
+}
+
+object CanBuildSingleArgumentFunction {
+  implicit def canBuildExponential =
```
dlwh: implicit object is probably better

math/src/main/scala/breeze/symbolic/CanBuildSingleArgumentFunction.scala
```diff
@@ -0,0 +1,18 @@
...
+    new CanBuildSingleArgumentFunction[Exponential] {
+      override def build[I <: SymbolicFunction](i: I): Exponential[I] = new Exponential(i)
+    }
+
+  implicit def canBuildLogarithm =
```
dlwh: same

math/src/main/scala/breeze/symbolic/CanCreateFunction.scala
```diff
@@ -0,0 +1,167 @@
...
+  * @tparam F The type of the SymbolicFunction from which the Scala function is created
+  */
+trait CanCreateFunction[I, O, F <: SymbolicFunction] {
+  def createFunction(f: F): I => O
+}
+
+trait LowPriorityCanCreateFunction {
+  implicit def canCreateVectorizedSymbolicFunction[I, O, F <: SymbolicFunction](
+    implicit canCreateFunction: CanCreateFunction[I, O, F],
+    bf: breeze.linalg.support.CanMapKeyValuePairs[DenseVector[I], Int, I, O, DenseVector[O]]
```
dlwh: this is going to be highly inefficient in general. maybe that's fine?

math/src/main/scala/breeze/symbolic/CanCreateFunction.scala
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+  implicit def canCreateConstantOneFunction[I, O](
+    implicit zero: CanCreateZerosLike[I, O], exp: numerics.exp.Impl[O, O]) =
+    new CanCreateFunction[I, O, ConstOne] {
+      def createFunction(f: ConstOne) =
+        (i: I) => exp(zero(i))
+    }
+
+  implicit def canCreateVarFunction[T] = new CanCreateFunction[T, T, Var] {
```
dlwh: (in general, all implicits should have explicit return types)

math/src/main/scala/breeze/symbolic/CanDerive.scala
```diff
@@ -0,0 +1,130 @@
...
+object CanDerive extends LowPriorityCanDerive {
+  def apply[F <: SymbolicFunction](implicit derivable: CanDerive[F]) = derivable
+  type Aux[F <: SymbolicFunction, D0 <: SymbolicFunction] = CanDerive[F] { type D = D0 }
+  case class SingletonOf[T, U <: { type D }](widen: T { type D = U#D })
+  object SingletonOf {
+    implicit def mkSingleton[T <: {type D}](implicit t: T) =
+      SingletonOf[T, t.type](t)
+  }
+  implicit class RichDerivable[F <: SymbolicFunction](f: F) {
+    def derivation(implicit canDerive: CanDerive[F]): canDerive.D = canDerive.derivation(f)
```
dlwh: why not just put this method on SymbolicFunction?

math/src/main/scala/breeze/symbolic/SymbolicFunction.scala
```diff
@@ -0,0 +1,102 @@
...
+import shapeless._
+
+import scala.util.matching.Regex
+
+/**
+  * Base trait for all symbolic functions.
+  * Not sealed because we want to allow users to
+  * create their own subclasses.
+  */
+trait SymbolicFunction
```
dlwh: inheriting from NumericOps[SymbolicFunction] and using UFuncs for operator implementations seems like it would be good

math/src/main/scala/breeze/symbolic/SymbolicFunction.scala
```diff
@@ -0,0 +1,102 @@
...
+
+case class Var() extends SymbolicFunction {
```
dlwh: case object?

dlwh: how do you deal with multiple variables/partial derivatives/gradients?

math/src/main/scala/breeze/symbolic/SymbolicFunction.scala
```diff
@@ -0,0 +1,102 @@
...
+case class Const[T](const: T) extends ConstBase {
+  override def toString = const.toString
+}
+
+/*
+  * Having special ConstZero and ConstOne classes
+  * allows us to retain compile-time information on zeroness
+  * which can be used to e.g. simplify functions.
+  */
+case class ConstZero() extends ConstBase {
```
dlwh: case object?

math/src/main/scala/breeze/symbolic/SymbolicFunction.scala
```diff
@@ -0,0 +1,102 @@
...
+  override def toString = "0.00"
```
dlwh: why not just 0.0.toString?

math/src/main/scala/breeze/symbolic/SymbolicFunction.scala
```diff
@@ -0,0 +1,102 @@
...
+}
+
+case class ConstOne() extends ConstBase {
+  override def toString = "1.00"
```
dlwh: similarly

math/src/main/scala/breeze/symbolic/CanSimplify.scala
```diff
@@ -0,0 +1,252 @@
...
+  implicit def canSimplifyLogConst[T](log: numerics.log.Impl[T, T]) =
+    new CanSimplify[Logarithm[Const[T]]] {
+      type R = Const[T]
+      def simplify(f: Logarithm[Const[T]]): R = Const(log(f.fn.const))
+    }
+}
+
+object CanSimplify extends HighPriorityCanSimplify {
+  implicit class RichSimplifiable[F](f: F) {
+    def simplify(implicit canSimplify: CanSimplify[F]): canSimplify.R =
```
dlwh: i've been leaning towards UFuncs for this rather than CanX traits

math/src/main/scala/breeze/symbolic/CanCreateFunction.scala
nightscape: This is the fallback if no other conversion options can be found. Do you have an idea on what we could improve in the general case?

math/src/main/scala/breeze/symbolic/CanDerive.scala
nightscape: You mean like this?
```
trait SymbolicFunction {
  def derivation(implicit canDerive: CanDerive[this.type]): canDerive.D = canDerive.derivation(this)
}
```
That would be an option. I just followed the pattern I saw in other type-class examples where the data classes (like SymbolicFunction) are relatively dumb and don't know about the type-class concerns (like derivation).

I was even considering not restricting the type-classes on SymbolicFunction at all. That would allow users to plugin any "description of a function", even if the implementation is not under their control (a third-party library), given they can provide a way to derive/simplify/functionize. I only put the SymbolicFunction restriction in to facilitate debugging type errors.
But I can definitely change this if you prefer.

math/src/main/scala/breeze/symbolic/CanSimplify.scala
nightscape: I can give this a try. I've been only thinking of UFuncs as a way to abstract over arity, but it might as well work nicely here :+1:
Would you do this for all of vectorization, derivation, simplification and functionization?

math/src/main/scala/breeze/symbolic/SymbolicFunction.scala
nightscape: Currently not at all. In our case we only have one free variable. One could give Var a parameter which specifies the variable, although this would probably have to be a singleton in order for the compiler to be able to incorporate the information about different variables into the calculations.
But that is currently beyond the scope of what we need / can invest time to do...

math/src/main/scala/breeze/symbolic/SymbolicFunction.scala
nightscape: I wanted to be able to distinguish it from a `Const[Double](0.0)` because that's important information when trying to find out why e.g. a formula doesn't get simplified.

nightscape:
Hi David,

I played around with the code a bit, made `SymbolicFunction` extend `NumericOps` and added `UFunc`s for derivation (`derive`), simplification (`simplify`), vectorization (`vectorize`) and functionization (`toFunction`). The `toFunction` `UFunc` is a little awkward to call, so I guess that the more common scenario will be to use the provided implicit wrapper like this
```scala
import CanCreateFunction._
Const(3.0).toFunction[Double, Double]
```

I also changed several `implicit def`s to `implicit val`s where no type parameter or implicit argument was needed. I went for `implicit def` + `implicit val` because it feels more consistent than `implicit def` + `implicit object` and [doesn't seem to have much of an impact otherwise](http://stackoverflow.com/questions/33317974/scala-implicit-object-vs-implicit-val).

Regarding tests: Would you go for run-time or compile-time tests? Does breeze already have compile-time tests from which I could borrow?

Also, it would be nice to make the compiler perform simplifications until it reaches a fix-point, but I have no idea how to do that... Currently you have to call `simplify` multiple times.

nightscape:
Hi @dlwh,

we've changed the approach rather radically: Instead of trying to make the Scala compiler do all the symbolic computation, we wrote macros that transform `SymbolicFunction` into an AST of the [Symja](https://bitbucket.org/axelclk/symja_android_library/overview) library, let Symja do the derivation / simplification and transform the result into a `CanDerive` / `CanSimplify` instance.
This gives us much better simplifications and could probably also be used to do symbolic computations other than deriving and simplifying.
Symja has a dual-license model, where the source code is LGPL, but the byte-code is Apache 2.
AFAIS this should be ok as we're only using a JAR dependency, but if you see a potential for problems, I can contact the author and ask for explicit permission / clarification of this use case.

I've also started adding some tests, so this is getting serious :smile: 
It would be great if you could have another look if this is heading in the right direction.

PR #609, opened 2016-12-01
Show correct cosineDistance formula in scaladoc

zakpatterson:
The documented behavior of this function is that it returns cosine(a, b). The cosine is (a dot b)/(norm(a) * norm(b)). This is 1.0 if the two vectors are equal, and 0.0 if they are orthogonal.  cosineDistance(a, b) = 1 - cosine(a, b).  

https://reference.wolfram.com/language/ref/CosineDistance.html


dlwh:
thanks!

PR #612, opened 2016-12-12
slight bugfix

tsobieszek:
lack of (implicit rand: RandBasis=Rand) causes defalt Rand being used in Gamma and Poisson distributions.


dlwh:
thanks!

PR #617, opened 2017-01-12
Gaussian.probability returns negative probability

mou7:
The function  `Gaussian.probability(x, y)` returns a negative probability when x is larger than y. 

This comes from the fact that the probability P[x =< Z <= y], computed as P[Z <= y] - P[Z <= x],  is defined only if x <= y. 

I added a `require` statement that checks if the arguments given are valid.


mou7:
How can I make this 'travis-ci' happy ?

dlwh:
It's not you, it's me. I need to stabilize the build (tests are randomize and tolerances aren't perfect.)

PR #623, opened 2017-01-24
Use computeIfAbsent for efficiency in HashIndex. 

adampauls:


adampauls:
Apparently not ready to abandon Java < 8 yet.

PR #625, opened 2017-02-10
adds logit (inverse sigmoid) function

maxthomas:
Pretty simple change - just a quick implementation of logit - figured I'd throw up a MR. Let me know if there's more I can do, if desired, etc.


maxthomas:
not sure what's up with CI - successes seems non-deterministic.

dlwh:
thanks, sorry for delay

PR #630, opened 2017-03-20
Fixed two issues with hashCode/equals in SparseVector

superbobry:
1. `SparseVector` defined no `hashCode` and therefore violated the hashCode/equals contract.
2. sparse/sparse comparisons worked in O(size) instead of O(activeSize).

### DIFF
--- a/math/src/main/scala/breeze/linalg/SparseVector.scala
+++ b/math/src/main/scala/breeze/linalg/SparseVector.scala
@@ -90,7 +90,8 @@ class SparseVector[@spec(Double, Int, Float, Long) V](val array: SparseArray[V])
   /** This is always assumed to be equal to 0, for now. */
   def default: V = zero.zero
 
-  override def equals(p1: Any) = p1 match {
+  override def equals(other: Any) = other match {
+    case x: SparseVector[_] => this.array == x.array
     case x: Vector[_] =>
         this.length == x.length &&
           (valuesIterator sameElements x.valuesIterator)
--- a/math/src/test/scala/breeze/collection/mutable/SparseArrayTest.scala
+++ b/math/src/test/scala/breeze/collection/mutable/SparseArrayTest.scala
@@ -16,6 +16,7 @@ package breeze.collection.mutable
  limitations under the License.
 */
 
+import breeze.linalg.SparseVector
 import org.scalatest._
 import org.scalatest.junit._
 import org.scalatest.prop._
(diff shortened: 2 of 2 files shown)


math/src/main/scala/breeze/collection/mutable/SparseArray.scala
```diff
@@ -424,6 +424,24 @@ final class SparseArray[@specialized(Double, Int, Float, Long) V](var index: Arr
...
+          this.default == x.default &&
+          (valuesIterator sameElements x.valuesIterator)
+    case _ => false
+  }
+
+  override def hashCode(): Int = {
+    val indexHash = util.Arrays.hashCode(index)
+    val dataHash = ArrayUtil.zeroSkippingHashCode(data, 0, 1, used)
+    Seq(indexHash, dataHash, used, size, default).map(_.hashCode())
+        .foldLeft(0)((a, b) => 31 * a + b)
```
nikdon: Probably there is a point to move `_.hashCode()` inside `foldLeft` to avoid usage of `map`?

superbobry: Good point. This is just the auto-generated version form IntelliJ. Note that it'd always allocate a `Seq`, do you think this is okay? I can do a non-allocating but slightly uglier version w/o foldLeft.

nikdon: Yes, you gave a good note, and I think it's better to avoid unnecessary allocations. It will not be so ugly anyway :)

superbobry: Will do 👌 .

superbobry:
The tests failures seem to be unrelated to my commit. Is this normal?

dlwh:
yeah, tests have been flaky for a while and I haven't had time to figure them out.

sorry for slow response.

math/src/main/scala/breeze/linalg/SparseVector.scala
```diff
@@ -90,13 +90,16 @@ class SparseVector[@spec(Double, Int, Float, Long) V](val array: SparseArray[V])
...
-  override def equals(p1: Any) = p1 match {
+  override def equals(other: Any) = other match {
+    case x: SparseVector[_] => this.array == x.array
     case x: Vector[_] =>
         this.length == x.length &&
           (valuesIterator sameElements x.valuesIterator)
     case _ => false
   }
 
+  override def hashCode(): Int = array.hashCode()
```
dlwh: this breaks hashcode/equality contract between SparseVector and DenseVector

dlwh: i swore I implemented hashcode for SparseVector before

superbobry: The contract is that the hashCode is the same given the same content?

dlwh: yeah, DenseVector/HashVector/SparseVector can all be equals. For DenseVector, I defined https://github.com/scalanlp/breeze/blob/5fd292608f3152bb0126694f7454890fa47c3e78/math/src/main/scala/breeze/util/ArrayUtil.scala#L313 specifically to make this work quickly for DenseVector, and I defined hashcode for all tensors in https://github.com/scalanlp/breeze/blob/master/math/src/main/scala/breeze/linalg/Tensor.scala#L89 that should be getting picked up by SparseVector

superbobry: This is unfortunate because in that case `hashCode` for SparseVector is linear in terms of `size` instead of `activeSize`.

dlwh: Why? it's iterating over active pairs, so it's only non-zeros. It's definitely not as efficient as it could be since it's always boxed, but it's definitely linear in nnz...

superbobry: Sorry, my bad. Your implementation should indeed be OK, I'll update the PR.

superbobry:
Force-pushed the rebased version. Thanks for the review @nikdon @dlwh!

Issue #572, opened 2016-08-04
LBFGS-B return wrong result if we choose bad initial value


yanboliang:
If we choose the initial value which is far from the correct result, `LBFGS-B` will return wrong result and output error logs. You can refer the following code to reproduce this bug. This function has minimum `f(x)=2.0` when `x=1.0`. But if we choose the initial value far from 1.0, it will output wrong result.

```
    val solver = new LBFGSB(DenseVector[Double](1E-12), DenseVector[Double](Double.MaxValue))

    val f = new DiffFunction[DenseVector[Double]] {
      override def calculate(x: DenseVector[Double]): (Double, DenseVector[Double]) = {
        val cost = x(0) + 1.0/x(0)
        val grad = DenseVector(1.0 - 1.0/(x(0)*x(0)))
        (cost, grad)
      }
    }

    val nearX0 = DenseVector[Double](1.5)
    val nearRes = solver.minimizeAndReturnState(f, nearX0)
    println(nearRes.x)
    println(nearRes.value)

    val farX0 = DenseVector[Double](1500)
    val farRes = solver.minimizeAndReturnState(f, farX0)
    println(farRes.x)
    println(farRes.value)
```

The output is:

```
DenseVector(1.0000013234582712)
2.0000000000017515
ERROR breeze.optimize.LBFGSB - Failure! Resetting history: breeze.optimize.FirstOrderException: Line search failed
ERROR breeze.optimize.LBFGSB - Failure again! Giving up and returning. Maybe the objective is just poorly behaved?
DenseVector(1500.0)
1500.0006666666666
```

yanboliang:
cc @dlwh

debasish83:
@yanboliang did you try NonlinearMinimizer with projection on bound ? NonlinearMinimizer has ADMM solver as well that supports bound but in my experiments NonlinearMinimizer with project on bound converges faster...Use NonlinearMinimizer.project and pass the proximal for the bound...

val lb = DenseVector.zeros[Double](ndim)
val ub = DenseVector.ones[Double](ndim)
val solver = NonlinearMinimizer.project(ProjectBox(lb, ub))

After that it works like FirstOrderMinimizer...solver.minimizeAndReturnState(huberLoss, init)

I will look into the LBFGS-B error...

yanboliang:
#633 fix this issue, so close it. Thanks!

PR #633, opened 2017-03-31
Fix lbfgsb linesearch out of bound and findAlpha method

WeichenXu123:
# What to solve
After a deep check to the LBFGS-B in breeze, I found two serious bug in LBFGS-B:
- line search is wrong in this implementation. 
- `LBFGSB.findAlpha` method is also wrong.

# Fix Line search with bound
According to the LBFGS-B paper
http://users.iems.northwestern.edu/~nocedal/PDFfiles/limited.pdf
The line search in LBFGS-B should be restricted in the bound, but the implementation in breeze do not,
this will cause the optimizer run out of bound, it will cause wrong result or cause line search fail, in some case.
This is the root cause of issue #572, and a series of features is blocking because of this bug:  
- Huber loss: https://github.com/apache/spark/pull/14326
- Bounded LiR: https://github.com/apache/spark/pull/17360
- Bounded LoR: https://issues.apache.org/jira/browse/SPARK-20047

So that it should be fixed ASAP.  
cc @yanboliang 

## Strong wolfe line search with bound restriction
As mentioned in paper, the best linesearch method for LBFGS-B is strong wolfe line search with bound restriction, this require some modification based on `StrongWolfeLineSearch` in breeze:

### modified strong wolfe condition with bound
We know that strong wolfe condition is to satisfy **sufficient decrease condition** and **curvature condition**, BUT according to the paper, the strong wolfe line search in LBFGS-B, the condition should be modified as following:
- satisfy **sufficient decrease condition**
- satisfy **curvature condition** OR **new point X(k + 1) hit the bound**

### Algorithm for strong wolfe condition with bound
Without bound, we already have the following algos (Nodecal & Wright Numerical Optimization p58)
lbfgsb-1
So that, with bound, we can modify this algo into following:
lbfgsb-2

## modification in LBFGS-B
- In method `determineStepSize`, first calculate the max step size which won't exceed bound box, then use it as the step bound, call `StrongWolfeLineSearch.minimizeWithBound`
- override `takeStep` method, in this method check whether the `newX` exceed bound and correct it. (This is used to avoid numerical error that cause the `newX` run out of bound)

# Fix `findAlpha` method
Unfortunately, the `findAlpha` method here is also wrong, the wrong implementation cause the `subspaceMin` point walk out of bound and so that in some case it will also cause the algos crash.
I trace to the code here through several failed testcases with `Huber loss`.
In summary, there are at least 2 mistakes in this method, let me explain what the method should do first:

find the maximum `alpha`, satisfiy `0.0 <= alpha <= 1.0`, and, for each dimension `i`, should satisfy:
```
lowerBound_i - xCauchy_i <= alpha * du_i <= upperBound_i - xCauchy_i
```

`xCauchy` is the Cauchy point inside the bound box which will satisfy
```
lowerBound_i - xCauchy_i <= 0
upperBound_i - xCauchy_i >= 0
```
and `du` is the direction vector (details please refer to the paper), the key point is ,  **for each `i`, the condition above should be satisfied**, so that the `subspace minimum point` computed will be restricted in the bound box.

So that the algo to find the maximum alpha should be:
```
minimize i: 0 -> dimensionSize  {
  if (du_i < 0) (lowerBound_i - xCauchy_i) / du_i
  else if (du_i > 0) (upperBound_i - xCauchy_i) / du_i
  else Double.Infinity
}
```
Note that we should handle the case `du_i == 0` carefully, otherwise it may generate `NaN` in computation and will cause the whole algo crash.

Then we can check the implementation in breeze, the logic in `findAlpha` is wrong. The place where it should use `math.min` it use `math.max`. And the code write `(ub_i - xc_i) / du_i` to be the wrong code `ub_i - xc_i / du_i`

The second mistake in `findAlpha` is that it do not handle the case that components of `du` is zero. This may cause computation run into `NaN`. We should skip the zero components of `du`.

# Numerical error handling
In theory, the `cauchy point`, the `subspaceMin point`, the `X` point, in the computation, should all be restricted in the bound box. BUT because of floating point error, it may slightly exceed the bound, so I add a `adjustWithinBound` method to correct it. So that it can avoid the point step out of bound which may cause other bugs.

# Test
The following typical algos have been tested:
- huber loss regression
- bounded LOR
- bounded LIR

### DIFF
--- a/math/src/main/scala/breeze/optimize/LBFGSB.scala
+++ b/math/src/main/scala/breeze/optimize/LBFGSB.scala
@@ -77,14 +77,46 @@ class LBFGSB(lowerBounds: DenseVector[Double],
 
 
   override protected def determineStepSize(state: State, f: DiffFunction[DenseVector[Double]], direction: DenseVector[Double]): Double =  {
-    val x = state.x
-    val ff = LineSearch.functionFromSearchDirection(f, x, direction)
+    val ff = new DiffFunction[Double] {
+      def calculate(alpha: Double) = {
+        val newX = takeStep(state, direction, alpha)
+        val (ff, grad) = f.calculate(newX)
+        ff -> (grad dot direction)
+      }
+    }
     val wolfeRuleSearch = new StrongWolfeLineSearch(maxZoomIter, maxLineSearchIter) // TODO: Need good default values here.
-    wolfeRuleSearch.minimize(ff, 1.0)
+
+    var minStepBound = Double.PositiveInfinity
+    var i = 0
+    while (i < lowerBounds.length) {
+      val dir = direction(i)
+      if (dir != 0.0) {
+        val bound = if (dir < 0.0) lowerBounds(i) else upperBounds(i)
+        val stepBound = (bound - state.x(i)) / dir
+        assert(stepBound >= 0.0)
+        if (stepBound < minStepBound) {
+          minStepBound = stepBound
+        }
+      }
+      i += 1
+    }
+
+    wolfeRuleSearch.minimizeWithBound(ff, 1.0, minStepBound)
   }
 
   override protected def takeStep(state: State, dir: DenseVector[Double], stepSize: Double) = {
-    state.x + (dir *:* stepSize)
+    val newX = state.x + (dir :* stepSize)
+    var i = 0
+    while (i < newX.length) {
+      if (newX(i) > upperBounds(i)) {
+        newX(i) = upperBounds(i)
+      }
+      if (newX(i) < lowerBounds(i)) {
+        newX(i) = lowerBounds(i)
+      }
+      i += 1
+    }
+    newX
   }
 
   private def initialize(f: DiffFunction[DenseVector[Double]], x0: DenseVector[Double]) = {
--- a/math/src/main/scala/breeze/optimize/LBFGSB.scala
+++ b/math/src/main/scala/breeze/optimize/LBFGSB.scala
@@ -101,7 +101,8 @@ class LBFGSB(lowerBounds: DenseVector[Double],
       i += 1
     }
 
-    wolfeRuleSearch.minimizeWithBound(ff, 1.0, minStepBound)
+    val initStep = if (minStepBound < 1.0) minStepBound else 1.0
+    wolfeRuleSearch.minimizeWithBound(ff, initStep, minStepBound)
   }
 
   override protected def takeStep(state: State, dir: DenseVector[Double], stepSize: Double) = {
--- a/math/src/main/scala/breeze/optimize/LBFGSB.scala
+++ b/math/src/main/scala/breeze/optimize/LBFGSB.scala
@@ -17,9 +17,10 @@ package breeze.optimize
 
 import breeze.linalg.{DenseMatrix, DenseVector}
 import breeze.linalg._
-import breeze.optimize.FirstOrderMinimizer.{State, ProjectedStepConverged, ConvergenceCheck}
+import breeze.optimize.FirstOrderMinimizer.{ConvergenceCheck, ProjectedStepConverged, State}
 import breeze.util.SerializableLogging
 import breeze.util.Implicits._
+import spire.syntax.cfor._
 
 /**
  * This algorithm is refered the paper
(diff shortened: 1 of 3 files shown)


math/src/main/scala/breeze/optimize/StrongWolfe.scala
```diff
@@ -58,12 +58,18 @@ class StrongWolfeLineSearch(maxZoomIter: Int, maxLineSearchIter: Int) extends Cu
...
 
+  def minimize(f: DiffFunction[Double], init: Double = 1.0): Double = {
+    minimizeWithBound(f, init = 1.0, bound = Double.PositiveInfinity)
+  }
+
   /**
    * Performs a line search on the function f, returning a point satisfying
    * the Strong Wolfe conditions. Based on the line search detailed in
    * Nocedal & Wright Numerical Optimization p58.
    */
```
yanboliang: Should we update the annotation?

WeichenXu123:
cc @yanboliang

math/src/main/scala/breeze/optimize/LBFGSB.scala
```diff
@@ -77,14 +77,47 @@ class LBFGSB(lowerBounds: DenseVector[Double],
...
+
+    val initStep = if (minStepBound < 1.0) minStepBound else 1.0
+    wolfeRuleSearch.minimizeWithBound(ff, initStep, minStepBound)
   }
 
   override protected def takeStep(state: State, dir: DenseVector[Double], stepSize: Double) = {
-    state.x + (dir *:* stepSize)
+    val newX = state.x + (dir :* stepSize)
+    var i = 0
+    while (i < newX.length) {
```
dlwh: i prefer cforRange these days

dlwh:
looks great! thanks so much!

So sorry for the long delay. work and life have been extra busy

WeichenXu123:
cc @dlwh @yanboliang @dbtsai 
Thanks!

WeichenXu123:
cc @sethah
Would you also take a look if you know some math details in this ? :)

yanboliang:
I have run test for bound constraint LiR/LoR and huber regression against this fix, all passed. Thanks !

dbtsai:
LGTM. Ping @dlwh to cut a new release for our usage in Spark if this looks okay. Thanks.

dlwh:
thanks!

dlwh:
i'll cut a release this week

math/src/main/scala/breeze/optimize/StrongWolfe.scala
```diff
@@ -58,12 +58,18 @@ class StrongWolfeLineSearch(maxZoomIter: Int, maxLineSearchIter: Int) extends Cu
   val c1 = 1e-4
   val c2 = 0.9
 
+  def minimize(f: DiffFunction[Double], init: Double = 1.0): Double = {
+    minimizeWithBound(f, init = 1.0, bound = Double.PositiveInfinity)
```
nerh: LBFGS is passing init value not always equals to 1 to this method. It that right to ignore it here?

WeichenXu123: In LBFGS-B, we can use 1 as the init value, according to paper, or do you have some better init value ?

dlwh:
i think it's more that you're ignoring the passed in value of init

math/src/main/scala/breeze/optimize/StrongWolfe.scala
nerh: Personally, I don't have a better init value, but LBFGS has: https://github.com/scalanlp/breeze/blob/master/math/src/main/scala/breeze/optimize/LBFGS.scala#L76 :)

Let me describe my problem: I've updated spark in my project to latest version (2.2) and some test on logistic regression start failing.
Spark 2.2 uses Breeze 0.13.1 and when I'm explicitly downgrading it to 0.13 tests stop failing.
It seems like in my case regression could not be successfully trained using LBFGS with init value equal to 1.

WeichenXu123: oh! you're right. we should fix it. like:
```
def minimize(f: DiffFunction[Double], init: Double = 1.0): Double = {
  minimizeWithBound(f, init, bound = Double.PositiveInfinity)
}
```

WeichenXu123: Thanks for finding this bug!

PR #636, opened 2017-04-22
use JavaConverters instead of JavaConversions

qilab:
Replace JavaConversions with JavaConverters because JavaConversions is deprecated since Scala 2.12.
Some refactoring is also added.

### DIFF
--- a/math/src/main/scala/breeze/util/TopK.scala
+++ b/math/src/main/scala/breeze/util/TopK.scala
@@ -13,10 +13,10 @@
  See the License for the specific language governing permissions and
  limitations under the License.
 */
-package breeze.util;
+package breeze.util
 
-import java.util.TreeSet;
-import scala.collection.JavaConverters._;
+import java.util.TreeSet
+import scala.collection.JavaConverters._
 
 /**
  * A Top-K queue keeps a list of the top K elements seen so far as ordered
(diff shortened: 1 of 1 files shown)


math/src/main/scala/breeze/util/TopK.scala
```diff
@@ -13,74 +13,74 @@
...
  * by the given comparator.
  */
-class TopK[T](k : Int)(implicit ord : Ordering[T]) extends Iterable[T] {
+class TopK[T](k: Int)(implicit ord: Ordering[T]) extends Iterable[T] {
 
-  private val keys = new TreeSet[T](ord);
+  private val keys = new TreeSet[T](ord)
 
-  def +=(e : T) = {
+  def +=(e: T): this.type = {
```
qilab: The return type of this method was `AnyVal` because of `if` statement without `else`.
However, I think, more specific type will be better: `Boolean`, `Unit` or `this.type` in this case.
`keys.add(e)` returns `true` if `keys` did not already contain the element `e`, so `Boolean` (or `Unit`) is a candidate of the return type.
but many `+=` methods in Scala Standard Library returns themselves (`this.type`) so `this.type` will be better.
WDYT?

dlwh: looks great

dlwh:
this is super old code, thanks for modernizing it

PR #639, opened 2017-05-03
remove some deprecated warnings

qilab:
This PR removes following warnings:

* type Context in package macros is deprecated (since 2.11.0): use blackbox.Context or whitebox.Context instead
* method encoded in class NameApi is deprecated (since 2.11.0): use `encodedName.toString` instead
*  method newTypeName in trait Names is deprecated (since 2.11.0): use TypeName instead
* method newTermName in trait Names is deprecated (since 2.11.0): use TermName instead
* method typeCheck in trait Typers is deprecated (since 2.11.0): use `c.typecheck` instead
* method fresh in trait Names is deprecated (since 2.11.0): use freshName instead

### DIFF
--- a/macros/src/main/scala/breeze/macros/arityize.scala
+++ b/macros/src/main/scala/breeze/macros/arityize.scala
@@ -25,10 +25,10 @@ object arityize {
         val maxOrder: Int = extractOrder(c)
 
         val results = for(order <- 1 to maxOrder) yield {
-          val bindings = Map(name.encoded -> order)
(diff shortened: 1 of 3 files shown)


dlwh:
we still support 2.10 so this doesn't work. We can split into scala-2.11/scala-2.12 and scala-2.10 sources, but that's it. If/When Spark stops supporting 2.10, we'll stop

qilab:
I see.
Thanks for your review. This PR must be closed.

Issue #563, opened 2016-07-15
Provide container matrix to avoid excessive allocations


XapaJIaMnu:
Hey,

In our application we have a lot of `DenseVector * DenseVector.t` operations which produce a matrix (about 20% of our runtime in total).  We suspect that part of the problem is that every time those two vectors are multiplied to produce a matrix, a new `DenseMatrix` object is being created resulting in excessive memory allocations and hence a lot of GC work. Now we actually need to keep only one version of our the matrix so we want to do something like:
`val container = DenseMatrix.zeros[Double](i,j)
container := vector*vector.t
`
Where the container matrix is passed to the function doing the `vector*vector.t`
However we don't see the expected decrease in the runtime or number of objects created in the profiler. I suspect that an intermediate Matrix is still being created. Is that the case?  Could we avoid it?

Cheers,

Nick

dlwh:
that's right. Breeze right now is fairly literal and doesn't have the
capacity to do fairly obvious optimizations. It's one of the many things
I'd like to add, but I don't have the bandwidth for it right now. You could
drop down and directly call the blas routines for outer product (dger or
sger) if you wanted.

On Fri, Jul 15, 2016 at 8:28 AM, Nikolay Bogoychev <notifications@github.com

dlwh:
dupe of #226

PR #648, opened 2017-06-26
Lazily populate dense vectors. 

stucchio:
I've been running into issues with code generating too many dense matrices, many of which are not necessary. 

For example, consider this issue: https://github.com/scalanlp/breeze/issues/563 The issue is basically that taking the outer product of two vectors results in storing an NxN matrix (immediately discarded) instead of two N-dimensional vectors. (I've also run into this issue.)

Similarly, an Identity matrix actually requires no storage at all. 

This PR allows a DenseMatrix to be constructed with a `MatrixPopulator` - a class that will construct the data storage only when (if) needed. 

Eventually I'd like to do things like, e.g., resolving #563 by building an `OuterProductMatrixPopulator` and similar things. But since this is a relatively big change, I figured I'd request feedback before doing too much.

### DIFF
--- a/math/src/main/scala/breeze/linalg/DenseMatrix.scala
+++ b/math/src/main/scala/breeze/linalg/DenseMatrix.scala
@@ -98,6 +98,22 @@ abstract class FunctionMatrixPopulator[V](val rows: Int, val cols: Int)(implicit
   }
 }
 
+abstract class DiagonalMatrixPopulator[V:ClassTag:Zero](dim: Int) extends MatrixPopulator[V] {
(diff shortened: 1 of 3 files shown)


dlwh:
Sorry, I meant to respond to this a long time ago. I think this isn't the right direction for this kind of functionality, and it's a huge change to a pretty fundamental type. I'd rather see something more like Lazy[DenseVector[T]] or Expr[DenseVector[T]] or some other wrapper. (I like the latter more because it starts to open up the ability to do more operation fusion...)

PR #654, opened 2017-09-01
Add overflow check in SVD

WeichenXu123:
The `workSize` in SVD is possible overflow to be negative value (when `m` and `n` is large), it will cause the algo throw `NegativeArraySizeException`. This exception will confuse user.

Add a check for it and if `m` and `n` is too large, throw a user-friendly exception message.

We have done this in spark(https://github.com/apache/spark/pull/19078) but it is better to fix it in breeze.


WeichenXu123:
cc @dlwh Thanks!

dlwh:
nice catch!

PR #658, opened 2017-09-19
Changed 'VectorBuilder.clear' to work in O(1)

superbobry:
This commit also removes redundant allocations from 'VectorBuilder.compact'.

### DIFF
--- a/math/src/main/scala/breeze/linalg/VectorBuilder.scala
+++ b/math/src/main/scala/breeze/linalg/VectorBuilder.scala
@@ -112,8 +112,7 @@ class VectorBuilder[@spec(Double, Int, Float, Long) E](private var _index: Array
     boundsCheck(i)
 
     if(_data.length <= used) {
-      _data = ArrayUtil.copyOf(_data, math.max(_data.length * 2, 1))
-      _index = ArrayUtil.copyOf(_index, math.max(_index.length * 2, 1))
+      reallocate(math.max(_data.length * 2, 1))
     }
 
     _data(used) = v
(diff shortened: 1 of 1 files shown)


math/src/main/scala/breeze/linalg/VectorBuilder.scala
```diff
@@ -241,22 +244,11 @@ class VectorBuilder[@spec(Double, Int, Float, Long) E](private var _index: Array
...
-    clear()
-    reserve(ah.activeSize)
-    var i = 0
-    while(i < ah.iterableSize) {
-      if(ah.isActive(i)) {
-        add(ah.index(i), ah.data(i))
-      }
-      i += 1
-    }
+    reallocate(used)
```
dlwh: `compact` is meant to remove "0" values and duplicates, so this isn't right. (Obviously, I should document this.)

math/src/main/scala/breeze/linalg/VectorBuilder.scala
```diff
@@ -241,22 +244,11 @@ class VectorBuilder[@spec(Double, Int, Float, Long) E](private var _index: Array
...
   }
 
   def clear() {
     used = 0
-    _index = new Array[Int](0)
```
dlwh: I'd rather clear actually release the arrays? Do you think that's not ideal?

superbobry: Unless I'm mistaken, the benefits of clearing/reusing the builder with the current "release the arrays" semantics are marginal. 

Imagine building tens of thousands of vectors with 100-200 elements. Since you know the maximum size in advance, you can set `initialNonzero` to a reasonable value to minimize reallocations. Then you build the vectors sequentially one after another, calling `clear` in between. After the `clear` call, the sizes of `_index` and `_data` are zero, which means that you'd have to do 7-8 reallocations for all but the first vector built.

superbobry:
Hey @dlwh, no worries. I'll address the issue in `compact`. Looking forward to your thoughts on `clear`.

math/src/main/scala/breeze/linalg/VectorBuilder.scala
dlwh: ok fair enough, i'm sold

PR #659, opened 2017-09-27
Changed 'Field' instances not to use 'Rich*.abs'

superbobry:
This is to workaround boxing on Scala 2.10, see SI-7511 and scala/scala#2676.


dlwh:
thanks!

PR #660, opened 2017-10-10
Update to SBT 1.0

nightscape:
Some plugins required updates as well, jacoco4sbt was replaced by a more modern
alternative and the IDE plugins removed, as those should go to the users global
plugins.sbt.


dlwh:
thanks!

PR #672, opened 2017-12-18
Change FirstOrderMinimizer to return convergence reason

mrkm4ntr:
I want to implement early-stopping using my own convergenceCheck implementation. If the loss of validation data increase, the value of the previous iteration is more appropriate.
So I changed FirstOrderMinimizer to return convergence reason.


dlwh:
thanks!

PR #675, opened 2017-12-21
Generalized OWLQN constructor to accept ConvergenceCheck

superbobry:
This makes the API equivalent to that of LBFGS.


dlwh:
thanks!

superbobry:
That was fast :)

dlwh:
caught me while doing email :)

PR #681, opened 2018-01-11
decomposition of LAPACK LU return values

ZacBlanco:
I probably still need to add some unit tests (I'm not terribly familiar with the project). I'm also not a Scala expert - but this is what I imagined the API to be like. If you have any suggestions I'm obviously willing to make changes.

### DIFF
--- a/math/src/main/scala/breeze/linalg/functions/LU.scala
+++ b/math/src/main/scala/breeze/linalg/functions/LU.scala
@@ -91,9 +91,6 @@ object LU extends UFunc {
     *         size RxR
      */
   def createPermutationMatrix(ipiv: Array[Int], rows:Int, cols: Int): DenseMatrix[Double] = {
-    val ipiv_dim: Int = scala.math.max(rows, cols)
-
-    // Create permutation matrix
     val indices: Array[Int] = new Array[Int](rows)
     for(i <- 0 until rows) {
       indices(i) = i
--- a/math/src/main/scala/breeze/linalg/functions/LU.scala
+++ b/math/src/main/scala/breeze/linalg/functions/LU.scala
@@ -98,7 +98,7 @@ object LU extends UFunc {
     }
 
     // Create permutation matrix
-    var pm: BDM[Double] = BDM.eye[Double](rows)
+    var pm: DenseMatrix[Double] = DenseMatrix.eye[Double](rows)
     for( r1 <- ipiv_dim - 1 to 0 by -1) {
       val r2: Int = ipiv(r1) - 1
       if (r1 != r2) {
@@ -132,7 +132,7 @@ object LU extends UFunc {
     * Differs from Breeze's currently implemented method in that it works on non-square matrices
    */
   private def lowerTriangular(X: DenseMatrix[Double]): DenseMatrix[Double] = {
-    BDM.tabulate(X.rows, X.cols)( (i, j) =>
+    DenseMatrix.tabulate(X.rows, X.cols)( (i, j) =>
       if (j == i) {
         1
       }else if (j < i) {
@@ -148,7 +148,7 @@ object LU extends UFunc {
     * Differs from Breeze's currently implemented method in that it works on non-square matrices
    */
   private def upperTriangular(X: DenseMatrix[Double]): DenseMatrix[Double] = {
-    BDM.tabulate(X.rows, X.cols)( (i, j) =>
+    DenseMatrix.tabulate(X.rows, X.cols)( (i, j) =>
       if (j == i) {
         X(i, j) // Keep values on the diagonal (unlike in lower triangular)
       } else if (j > i) {
--- a/math/src/main/scala/breeze/linalg/functions/LU.scala
+++ b/math/src/main/scala/breeze/linalg/functions/LU.scala
@@ -4,7 +4,6 @@ import org.netlib.util.intW
 import com.github.fommil.netlib.LAPACK.{getInstance => lapack}
 import breeze.generic.UFunc
 import breeze.math.Semiring
-
 /**
  * Computes the LU factorization of the given real M-by-N matrix X such that
  * X = P * L * U where P is a permutation matrix (row exchanges).
@@ -81,31 +80,20 @@ object LU extends UFunc {
     * @param ipiv - The pivot vector returned from LAPACK
     * @param rows - The number of rows in the original matrix
     * @param cols - The number of columns in the original matrix
-    * @returns The permutation matrix, P from the LU decomposition of the form (P * L * U) 
+    * @return The permutation matrix, P from the LU decomposition of the form (P * L * U)
      */
-  private def createPermutationMatrix(ipiv: Array[Int], rows:Int, cols: Int): DenseMatrix[Double] = {
-    val ipiv_dim: Int = min(rows, cols).toInt
-    var parr: Array[Int] = new Array(rows)
-
-    // Create perm vector
-    for (x <- 0 to rows-1) {
-      parr(x) = x + 1
-    }
-    for(i <- 0 to ipiv_dim-1) {
-      val i2: Int = ipiv(i) - 1
-      var tmp: Int = parr(i)
-      parr(i) = parr(i2)
-      parr(i2) = tmp
-    }
+  def createPermutationMatrix(ipiv: Array[Int], rows:Int, cols: Int): DenseMatrix[Double] = {
+    val ipiv_dim: Int = scala.math.min(rows, cols)
 
     // Create permutation matrix
-    var pm: DenseMatrix[Double] = DenseMatrix.eye[Double](rows)
-    for( r1 <- ipiv_dim - 1 to 0 by -1) {
+    val pm: DenseMatrix[Double] = DenseMatrix.eye[Double](rows)
+    var tmp: Double = 0.0
+    for ( r1 <- ipiv_dim - 1 to 0 by -1) {
       val r2: Int = ipiv(r1) - 1
       if (r1 != r2) {
-        for (col <- 0 to rows - 1) {
+        for (col <- 0 until rows) {
           // Swap pm(r1, col) with pm(row2, col)
-          val tmp: Double = pm(r1, col)
+          tmp = pm(r1, col)
           pm(r1, col) = pm(r2, col)
           pm(r2, col) = tmp
         }
@@ -122,7 +110,7 @@ object LU extends UFunc {
     * i.e X, ipiv = LU(A), P, L, U = decompose(X, ipiv), P * L * U = A
   */
   def decompose(X: DenseMatrix[Double], ipiv: Array[Int]): (DenseMatrix[Double], DenseMatrix[Double], DenseMatrix[Double]) = {
-    val P: DenseMatrix[Double] = createPermutationMatrix(ipiv, X.rows.toInt, X.cols.toInt)
+    val P: DenseMatrix[Double] = createPermutationMatrix(ipiv, X.rows, X.cols)
     val L: DenseMatrix[Double] = lowerTriangular(X)
     val U: DenseMatrix[Double] = upperTriangular(X)
     (P, L, U)
@@ -131,6 +119,7 @@ object LU extends UFunc {
   /** 
     * Returns the lower triangular matrix from X.
     * Differs from Breeze's currently implemented method in that it works on non-square matrices
+    * Returns 1 on the diagonal
    */
   private def lowerTriangular(X: DenseMatrix[Double]): DenseMatrix[Double] = {
     DenseMatrix.tabulate(X.rows, X.cols)( (i, j) =>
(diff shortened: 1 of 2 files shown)


math/src/main/scala/breeze/linalg/functions/LU.scala
```diff
@@ -73,4 +74,90 @@ object LU extends UFunc {
...
+    * @param rows - The number of rows in the original matrix
+    * @param cols - The number of columns in the original matrix
+    * @returns The permutation matrix, P from the LU decomposition of the form (P * L * U) 
+     */
+  private def createPermutationMatrix(ipiv: Array[Int], rows:Int, cols: Int): DenseMatrix[Double] = {
+    val ipiv_dim: Int = min(rows, cols).toInt
+    var parr: Array[Int] = new Array(rows)
+
+    // Create perm vector
+    for (x <- 0 to rows-1) {
```
nikdon: I would suggest to use `spire.syntax.cfor._` and replace fors with `cfor` ot `crange`

math/src/main/scala/breeze/linalg/functions/LU.scala
```diff
@@ -73,4 +74,90 @@ object LU extends UFunc {
...
+  // convert from the vector to the matrix, this is what I came up with.
+  /**
+    * Creates the permutation matrix from a pivot vector (result from LAPACK)
+    * @param ipiv - The pivot vector returned from LAPACK
+    * @param rows - The number of rows in the original matrix
+    * @param cols - The number of columns in the original matrix
+    * @returns The permutation matrix, P from the LU decomposition of the form (P * L * U) 
+     */
+  private def createPermutationMatrix(ipiv: Array[Int], rows:Int, cols: Int): DenseMatrix[Double] = {
+    val ipiv_dim: Int = min(rows, cols).toInt
```
nikdon: Probably here I could use simple `scala.math.min(rows, cols)` instead of `breeze`s `min`.  Anyway, `toInt` is not required here.

math/src/main/scala/breeze/linalg/functions/LU.scala
nikdon: And if you want to fill new array with values based on number of elements you could use something like `Array.tabulate(rows)(_ + 1)`

math/src/main/scala/breeze/linalg/functions/LU.scala
```diff
@@ -73,4 +74,90 @@ object LU extends UFunc {
...
+    var parr: Array[Int] = new Array(rows)
```
nikdon: Why do  you need `parr`? It seems like this array is not required anywhere in the function.

math/src/main/scala/breeze/linalg/functions/LU.scala
```diff
@@ -73,4 +74,90 @@ object LU extends UFunc {
...
+
+  /**
+    * Decomposes the output of the LAPACK LU decomposition
+    * @param X - The packed upper/lower matrices from LAPACK
+    * @param ipiv - The pivot vector from LAPACK
+    * @return (DenseMatrix, DenseMatrix, DenseMatrix) - Three matrices P, L, and U which give the original matrix when multiplied together
+    * i.e X, ipiv = LU(A), P, L, U = decompose(X, ipiv), P * L * U = A
+  */
+  def decompose(X: DenseMatrix[Double], ipiv: Array[Int]): (DenseMatrix[Double], DenseMatrix[Double], DenseMatrix[Double]) = {
+    val P: DenseMatrix[Double] = createPermutationMatrix(ipiv, X.rows.toInt, X.cols.toInt)
```
nikdon: rows and cols are already `Int`s, so it's not necessary to call `toInt`

math/src/main/scala/breeze/linalg/functions/LU.scala
```diff
@@ -73,4 +74,90 @@ object LU extends UFunc {
...
+    }
+    for(i <- 0 to ipiv_dim-1) {
+      val i2: Int = ipiv(i) - 1
+      var tmp: Int = parr(i)
+      parr(i) = parr(i2)
+      parr(i2) = tmp
+    }
+
+    // Create permutation matrix
+    var pm: DenseMatrix[Double] = DenseMatrix.eye[Double](rows)
```
nikdon: It might be a `val` as method `=` that you use inside the loop is `update` from the `DenseMatrix`.

math/src/main/scala/breeze/linalg/functions/LU.scala
```diff
@@ -73,4 +74,90 @@ object LU extends UFunc {
...
+    for( r1 <- ipiv_dim - 1 to 0 by -1) {
+      val r2: Int = ipiv(r1) - 1
+      if (r1 != r2) {
+        for (col <- 0 to rows - 1) {
+          // Swap pm(r1, col) with pm(row2, col)
+          val tmp: Double = pm(r1, col)
```
nikdon: Here you can approach the loop with storing a tmp value outside the loop. It could save some time as it's not required to allocate a new val every iteration.
```scala
var tmp: Double = 0.0
        for (col <- 0 until rows) {
          // Swap pm(r1, col) with pm(row2, col)
          tmp = pm(r1, col)
          pm(r1, col) = pm(r2, col)
          pm(r2, col) = tmp
        }
```

math/src/main/scala/breeze/linalg/functions/LU.scala
ZacBlanco: Can you elaborate on this? I'm not understanding

math/src/main/scala/breeze/linalg/functions/LU.scala
ZacBlanco: I tried using `cfor`s - it didn't really make a difference in the speedup. Did 150 runs on a ipiv of size 500 and the difference was < 3ms

ZacBlanco: This was the code I used


```scala
    cfor(ipiv_dim-1)(i => i >= 0, i => i - 1)(r1 => {
      val r2: Int = ipiv(r1) - 1
      if (r1 != r2) {
        cfor(0)(i => i < rows, i => i + 1)(col => {
          // Swap pm(r1, col) with pm(row2, col)
          tmp = pm(r1, col)
          pm(r1, col) = pm(r2, col)
          pm(r2, col) = tmp
        })
      }
    })
```

math/src/main/scala/breeze/linalg/functions/LU.scala
```diff
@@ -73,4 +73,81 @@ object LU extends UFunc {
     }
   }
 
+  // Note: Not sure if this method could be faster or not, but after hours of searching how to
```
dlwh: i think you can just do http://icl.cs.utk.edu/lapack-forum/viewtopic.php?f=2&t=874#p3168

math/src/main/scala/breeze/linalg/functions/LU.scala
```diff
@@ -73,4 +73,81 @@ object LU extends UFunc {
...
+      }
+    )
+  }
+
+  /** 
+    * Returns the upper triangular matrix from X.
+    * Differs from Breeze's currently implemented method in that it works on non-square matrices
+    * Returns current values on the diagonal.
+   */
+  private def upperTriangular(X: DenseMatrix[Double]): DenseMatrix[Double] = {
```
dlwh: this can just be`breeze.linalg.upperTriangular(X)`

math/src/main/scala/breeze/linalg/functions/LU.scala
```diff
@@ -73,4 +73,81 @@ object LU extends UFunc {
...
+    val U: DenseMatrix[Double] = upperTriangular(X)
+    (P, L, U)
+  }
+
+  /** 
+    * Returns the lower triangular matrix from X.
+    * Differs from Breeze's currently implemented method in that it works on non-square matrices
+    * Returns 1 on the diagonal
+   */
+  private def lowerTriangular(X: DenseMatrix[Double]): DenseMatrix[Double] = {
```
dlwh: i'd do 

```scala
val r = breeze.linalg.strictlyLowerTriangular(X)
diag(r) := 1.0
```

math/src/main/scala/breeze/linalg/functions/LU.scala
```diff
@@ -73,4 +73,81 @@ object LU extends UFunc {
...
+  }
+
+  /**
+    * Decomposes the output of the LAPACK LU decomposition
+    * @param X - The packed upper/lower matrices from LAPACK
+    * @param ipiv - The pivot vector from LAPACK
+    * @return (DenseMatrix, DenseMatrix, DenseMatrix) - Three matrices P, L, and U which give the original matrix when multiplied together
+    * i.e X, ipiv = LU(A), P, L, U = decompose(X, ipiv), P * L * U = A
+  */
+  def decompose(X: DenseMatrix[Double], ipiv: Array[Int]): (DenseMatrix[Double], DenseMatrix[Double], DenseMatrix[Double]) = {
```
dlwh: could you a case class rather than a tuple?

math/src/main/scala/breeze/linalg/functions/LU.scala
dlwh: this is unlikely to make any difference, except quite possibly be slower because it has to generate a capture because of the lambda.

dlwh: (the tmp is going to get stored in a register probably so it's basically free)

math/src/main/scala/breeze/linalg/functions/LU.scala
dlwh: yeah, i don't think this is a critical thing to make super-fast, though I do think you should use the single-loop version I linked to

math/src/main/scala/breeze/linalg/functions/LU.scala
ZacBlanco: When I was doing tests previously, upperTriangular failed for a few case of matrix sizes.

i.e.
20 x 20 and 20 x 40 work,
40 x 20 does not. (I get a java.lang.IndexOutOfBounds Exception)

This method also keeps the original size of the matrix. Breeze's currently implemented methods always return a square matrix (correct me if I'm wrong).

Same applies for the strictlyLowerTriangular

dlwh: fair enough

dlwh:
ok, great. If you could do the faster permutation algorithm, everything would be great!

ZacBlanco:
The forum post you linked was just the code for converting the `ipiv` vector from LAPACK form to Matlab form. Didn't have anything to do with converting it into a matrix.

ZacBlanco:
[Here's the implementation from JBLAS](https://github.com/mikiobraun/jblas/blob/master/src/main/java/org/jblas/util/Permutations.java#L99-L119). I'll see how it performs against what I currently have. If it's faster I'll replace mine.

ZacBlanco:
Okay, so the JBLAS one is significantly faster for larger matrices (50ms vs 650ms) on a 5000x5000 matrix. Below is the code I used to test.

JBLAS implementation
```scala
def jBLASCreatePermutationMatrix(ipiv: Array[Int], rows:Int, cols: Int): DenseMatrix[Double] = {
    val ipiv_dim: Int = scala.math.min(rows, cols)

    // Create permutation matrix
    val indices: Array[Int] = new Array[Int](ipiv_dim)
    for(i <- 0 until ipiv_dim) {
      indices(i) = i
    }

    for(i <- 0 until ipiv.length) {
      val j = ipiv(i) - 1
      val t = indices(i)
      indices(i) = indices(j)
      indices(j) = t
    }

    val pm: DenseMatrix[Double] = DenseMatrix.zeros[Double](ipiv_dim, ipiv_dim)
    for(i <- 0 until ipiv_dim) {
      pm(indices(i), i) = 1
    }
    pm
  }
```

Scalatest for timing
```scala
test("Permutation Matrix Test") {
    var r: Int = 5000
    var c: Int = 5000
    val a: DenseMatrix[Double] = DenseMatrix.rand(r, c)
    var (lum, ipiv) = LU(a)
    var lu(p, l, u) = LU.decompose(lum, ipiv)
    var tick = System.currentTimeMillis()
    val perm1 = LU.createPermutationMatrix(ipiv, r, c)
    var tock = System.currentTimeMillis()
    var diff = tock-tick
    println(s"Original $diff")
    tick = System.currentTimeMillis()
    val perm2 = LU.jBLASCreatePermutationMatrix(ipiv, r, c)
    tock = System.currentTimeMillis()
    diff = tock - tick
    println(s"JBLAS $diff")
    assert(norm(sum(perm2-perm1)) < 10E-15)

  }
```

ZacBlanco:
Also a side note. I made the case class lowercase so it didn't conflict with the object name - however this isn't consistent with what the QR decomposition uses (lower case file/object name, upper case case class). Not sure how much you care about the consistency in this case but I thought I would at least bring it to your attention.

dlwh:
thanks! I'll probably tweak the names while I do the UFunc stuff

dlwh:
ok thanks so much! I applied some tweaks and made your thing the default for the LU function, making an `LU.primitive` function that has the old semantics.

Thanks again!

PR #683, opened 2018-01-14
Added indirectSort

superbobry:
Indirect sorting allows to sort one or more arrays based on the values
in the key array. The current implementation generalizes indexSort with
one (unfortunate) drawback: each call to indexSort now requires two
allocations.

### DIFF
--- a/math/src/main/scala/breeze/linalg/CSCMatrix.scala
+++ b/math/src/main/scala/breeze/linalg/CSCMatrix.scala
@@ -479,8 +479,8 @@ object CSCMatrix extends MatrixConstructors[CSCMatrix]
       val nnz = indices.length
 
 
-      val _rows = if (rows >= 0) rows else indices.map(i => (i & 0xFFFFFFFFL).toInt).foldLeft(0)(_ max _) + 1
-      val _cols = if (cols >= 0) cols else indices.map(i => (i >> 32).toInt).foldLeft(0)(_ max _) + 1
+      val _rows = if (rows >= 0) rows else indices.map(rowFromIndex).foldLeft(0)(_ max _) + 1
+      val _cols = if (cols >= 0) cols else indices.map(colFromIndex).foldLeft(0)(_ max _) + 1
 
       val outCols = new Array[Int](_cols+1)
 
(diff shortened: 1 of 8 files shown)


superbobry:
@dlwh what do you think about the extra allocation in `indexSort`? Also, I've noticed that `expand` does not generate the unmangled specializations, should I add them manually (I haven't done it yet because that's 16 boilerplate methods).

math/src/test/scala/breeze/linalg/VectorBuilderTest.scala
```diff
@@ -23,8 +23,8 @@ class VectorBuilderTest extends FunSuite with Checkers {
   test("Basic VectorBuilder result tests") {
     val vb = VectorBuilder[Double](0.0, 1.0, 3.0)
 
-    assert(vb.toSparseVector === SparseVector[Double](0.0, 1.0, 3.0))
-    assert(vb.toHashVector === HashVector[Double](0.0, 1.0, 3.0))
+//    assert(vb.toSparseVector === SparseVector[Double](0.0, 1.0, 3.0))
```
dlwh: ??

math/src/main/scala/breeze/linalg/CSCMatrix.scala
```diff
@@ -549,7 +549,7 @@ object CSCMatrix extends MatrixConstructors[CSCMatrix]
 
     // returns indices of a lexicographic sort of the indices (columns major, rows minor)
     private def sortedIndices(indices: Array[Long]) = {
-      Sorting.indexSort(indices)
+      Sorting.indexSort(indices, 0, indices.length)
```
dlwh: you can actually switch this one to use your new indirectSort since in the end we just do the mutation

math/src/main/scala/breeze/util/Sorting.scala
```diff
@@ -17,44 +17,42 @@ object Sorting {
...
-    indexSort(elems, 0, elems.length)
+    val idx = ArrayUtil.range(0, length)
+    indirectSort_Double_Int(elems.clone(), idx, 0, elems.length)
+    idx
   }
 
   @expand
-  def indexSort[@expand.args(Int, Long, Float, Double) T](elems: Array[T], elemsOff: Int, elemsLength: Int): Array[Int] = {
-    val idx = ArrayUtil.range(0, elemsLength)
+  def indirectSort[@expand.args(Int, Long, Float, Double) K, @expand.args(Int, Long, Float, Double) E](
```
dlwh: I think it might be fine to make `E` unexpanded. It's more expensive than not (since scala has to generate a pattern match), but I think it's probably unlikely to be too noticeable in practice since the branch is so predictable

math/src/main/scala/breeze/linalg/CSCMatrix.scala
dlwh: same for VectorBuilder.toSparseVector

superbobry: Yup, I have a followup commit doing this.

math/src/main/scala/breeze/util/Sorting.scala
superbobry: Do you mean unexpanded but specialized?

math/src/test/scala/breeze/linalg/VectorBuilderTest.scala
superbobry: Sorry, debugging leftover.

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.0.2
+sbt.version=1.0.3
```
superbobry: This is here to fix IntelliJ import bug: https://youtrack.jetbrains.com/issue/SCL-12736

superbobry:
@dlwh we have "diverging implicit expansion for type" in `master`.

dlwh:
yeah, I need to deal with that. there's a bug/limitation in the divergence checker for 2.10 that breaks one test. I'll fix soon, but I'm not worried so long as the other two builds are passing.

math/src/main/scala/breeze/util/Sorting.scala
dlwh: i didn't, but that's a better idea :)

dlwh:
could i get you to make CSCMatrixBuilder use the indirect sort? otherwise looks great!

superbobry:
Yup, currently working on the builder for CSCMatrix. Will push shortly.

superbobry:
Actually, I'm wondering if it's worth keeping `indexSort` at all? The extra copy could be non-obvious to the users and it's just a special case of `indirectSort`. What do you think?

dlwh:
sounds good to me!

Issue #684, opened 2018-01-15
Scalafmt


superbobry:
Have you considered using Scalafmt to make formatting a non-issue?  

I did a dumb PR applying Scalafmt to the whole codebase to get an approximation of the end result. The config I used is just the one we use at work. Feel free to comment on things you like/dislike, I'll try to make Scalafmt produce the desired output.

dlwh:
thanks!

PR #685, opened 2018-01-15
[RFC] Scalafmt

superbobry:
See #684


.scalafmt.conf
```diff
@@ -0,0 +1,11 @@
+maxColumn = 100
```
dlwh: set this to 120 and then I'm happy

superbobry: Done!

dlwh:
thanks!

PR #691, opened 2018-02-02
convert quickselect algorithm to use 3-way partition

numberten:
This changes the quickselect algorithm to make use of a 3-way partition. This makes quickselection more robust in the face of Arrays containing many identical elements, which currently exhibits quadratic (O(n<sup>2</sup>)) behavior. See: https://en.wikipedia.org/wiki/Dutch_national_flag_problem


dlwh:
thanks!

PR #693, opened 2018-02-13
Don't insert default value in OpenHashArray

vincent-grosbois:
If we are inserting a new element, and that element is the default, then
we don't do anything and don't increase the load factor.

NB: default value can still be inserted if it is inserted after a non-
default value that was inserted with the same key. We could remove the
entry in that case but that would require more work

### DIFF
--- a/math/src/main/scala/breeze/collection/mutable/OpenAddressHashArray.scala
+++ b/math/src/main/scala/breeze/collection/mutable/OpenAddressHashArray.scala
@@ -94,15 +94,8 @@ final class OpenAddressHashArray[@specialized(Int, Float, Long, Double) V] priva
   final def update(i: Int, v: V) {
     if (i < 0 || i >= size) throw new IndexOutOfBoundsException(i + " is out of bounds for size " + size)
     val pos = locate(i)
-
-    if (_index(pos) == i) {
-      _data(pos) = v
-    }
-    else if (v == defaultValue.value) {
-      return
-    }
-    else {
-      _data(pos) = v
+    _data(pos) = v
+    if (_index(pos) != i && v != defaultValue.value) {
       load += 1
       if (load * 4 > _index.length * 3) {
         rehash()
--- a/math/src/main/scala/breeze/collection/mutable/OpenAddressHashArray.scala
+++ b/math/src/main/scala/breeze/collection/mutable/OpenAddressHashArray.scala
@@ -95,8 +95,14 @@ final class OpenAddressHashArray[@specialized(Int, Float, Long, Double) V] priva
     if (i < 0 || i >= size) throw new IndexOutOfBoundsException(i + " is out of bounds for size " + size)
     val pos = locate(i)
 
-    _data(pos) = v
-    if (_index(pos) != i && v != defaultValue.value) {
+    if (_index(pos) == i) {
+      _data(pos) = v
+    }
+    else if (v == defaultValue.value) {
+      return
+    }
+    else {
+      _data(pos) = v
       load += 1
       if (load * 4 > _index.length * 3) {
         rehash()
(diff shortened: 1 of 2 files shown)


math/src/main/scala/breeze/collection/mutable/OpenAddressHashArray.scala
```diff
@@ -94,8 +94,9 @@ final class OpenAddressHashArray[@specialized(Int, Float, Long, Double) V] priva
   final def update(i: Int, v: V) {
     if (i < 0 || i >= size) throw new IndexOutOfBoundsException(i + " is out of bounds for size " + size)
     val pos = locate(i)
+
     _data(pos) = v
-    if (_index(pos) != i) {
+    if (_index(pos) != i && v != defaultValue.value) {
```
superbobry: You need to do it before line #98, otherwise, it's too late.

math/src/main/scala/breeze/collection/mutable/OpenAddressHashArray.scala
```diff
@@ -94,8 +94,15 @@ final class OpenAddressHashArray[@specialized(Int, Float, Long, Double) V] priva
   final def update(i: Int, v: V) {
     if (i < 0 || i >= size) throw new IndexOutOfBoundsException(i + " is out of bounds for size " + size)
     val pos = locate(i)
-    _data(pos) = v
-    if (_index(pos) != i) {
+
+    if (_index(pos) == i) {
```
superbobry: Am I right that this is the "found" case? Actually, I start to think the previous version is correct and maybe even more readable than the new one...

math/src/main/scala/breeze/collection/mutable/OpenAddressHashArray.scala
```diff
@@ -94,8 +94,15 @@ final class OpenAddressHashArray[@specialized(Int, Float, Long, Double) V] priva
...
+      _data(pos) = v
+    }
+    else if (v == defaultValue.value) {
```
superbobry: Please use Scalafmt ;)

math/src/test/scala/breeze/collection/mutable/OpenAddressHashArrayTest.scala
```diff
@@ -73,4 +73,13 @@ class OpenAddressHashArrayTest extends FunSuite with Checkers {
     assert(oah(1) == -1.0)
   }
 
+  test("adding default value doesn't do anything") {
```
dlwh: please add a test case like:

```
oah(0) = 1.0
oah(0) = -1.0
assert(oah(0) == -1.0)
```

vincent-grosbois: >this isn't correct, is it? If there's already something at i that's not a default value, this fails to overwrite it.
sorry I don't get it, in that case you enter into the branch at line 98 and overwrite the previous value with v

vincent-grosbois:
Hi!
I can put sentinel values on delete, but then would there be any plan at some point to "purge" the array of all its sentinels (and possibly shrink down the array as well) ?
Without ever removing the sentinels, putting sentinels seems exactly the same as explicetely storing the default value in the array I think

math/src/main/scala/breeze/collection/mutable/OpenAddressHashArray.scala
superbobry: I think in this case `_index(pos) == i`, no?

dlwh:
oh yeah, that's what I get for code reviewing when i'm going to sleep.

re: deletions, i think we can just punt

Maybe add my test case anyway and then I'm happy. Thanks for doing this!

vincent-grosbois:
I reverted to the first version of the code and changed the test, I think you should be able to see the updated PR

dlwh:
Thanks! So sorry for the slow response. I appreciate the contribution!

Issue #590, opened 2016-10-03
DenseTensor?


Hydrotoast:
From the wiki, I see that there exists `DenseVector` and `DenseMatrix`; however, is there a `DenseTensor` data structure?

dlwh:
no. people have offered to add one about 4-5 times, but no one's followed
through yet.

stormy-ua:
+1 to the issue. I was thinking about building a simple deep learning algorithms with breeze, but the lack of the support for Tensor makes it infeasible.

PR #695, opened 2018-02-25
[WIP] [BREEZE-590] DenseTensor Implementation

sujithjay:
#590 
The idea of this patch is to implement a DenseTensor, as the title suggests. It is still a work in progress; however, I decided to raise a PR anyways to elicit comments. I am still very new to Breeze and could do with some guidance, and a thorough review from someone with experience in the codebase.

The primary design choice I made is to choose TensorN over Tensor3, Tensor4 etc. This does give away static type safety, but brings in flexibility in choosing dimensions of the Tensor.

I have in mind the following checklist to complete this patch in the current form:

- [ ] Implement UFuncs
- [ ] Improve the iterator performance (by adding a cache, probably?)
- [ ] Exhaustive tests

Please feel free to suggest more items to add to the checklist. Thank you.

### DIFF
--- a/math/src/main/scala/breeze/linalg/DenseTensorN.scala
+++ b/math/src/main/scala/breeze/linalg/DenseTensorN.scala
@@ -1,0 +1,133 @@
+package breeze.linalg
+
+import scala.collection.mutable
+import scala.reflect.ClassTag
+import scala.{specialized => spec}
+
+ /**
+  * A DenseTensorN is an N-dimensional tensor with all elements flattened into an array. It is
+  * column major unless isTranspose is true.
+  *
+  * @author sujithjay
+  * @param data The underlying data.
+  *             Column-major unless isTranspose is true.
+  *             Mutate at your own risk.
+  *             Note that this tensor may be a view of the data.
+  * @param shape
+  * @param offset
+  * @param stride
+  * @param isTranspose
+  */
+@SerialVersionUID(1L)
+final class DenseTensorN[@spec(Double, Int, Float, Long) V](
+  val data: Array[V],
+  val shape: IndexedSeq[Int],
+  val offset: Int,
+  val stride: Int,
+  val isTranspose: Boolean = false)
+  extends TensorN[V]
+  with TensorNLike[V, DenseTensorN[V]]
+  with Serializable {
+
+  def this(data: Array[V]) = this(data, IndexedSeq[Int](data.length), 0, 1)
+  def this(shape: IndexedSeq[Int])(implicit man: ClassTag[V]) = this(new Array[V](shape.product), shape, 0, 1)
+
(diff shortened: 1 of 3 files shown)


math/src/main/scala/breeze/linalg/DenseTensorN.scala
```diff
@@ -0,0 +1,133 @@
...
+  def this(data: Array[V]) = this(data, IndexedSeq[Int](data.length), 0, 1)
+  def this(shape: IndexedSeq[Int])(implicit man: ClassTag[V]) = this(new Array[V](shape.product), shape, 0, 1)
+
+  def apply(index: IndexedSeq[Int]): V = {
+    val isValidIndex = index
+      .zipWithIndex
+      .forall{ idx =>
+        idx._2 > -shape(idx._1) && idx._2 < shape(idx._1)
+      }
+    if(!isValidIndex) throw new IndexOutOfBoundsException("")
```
dlwh: isValidIndex should be extracted out as a method

math/src/main/scala/breeze/linalg/DenseTensorN.scala
```diff
@@ -0,0 +1,133 @@
...
+        idx._2 + shape(idx._1) * prev
+      }
+    }
+    else {
+      ndIndex.zipWithIndex.foldRight(0){(idx, prev) =>
+        idx._2 + shape(idx._1) * prev
+      }
+    }
+
+    logicalIndex * (stride + 1) + offset
```
dlwh: why is this `stride + 1`? 

In any event, I don't think that an element-by-element stride isn't really that useful for higher-order tensors. In practice, I think you'd you want a per-dimension stride. (Or, alternatively, the stride only to apply to the "major" axis, as in DenseMatrix)

math/src/main/scala/breeze/linalg/TensorN.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  def shape: IndexedSeq[Int]
+
+  def keySet: Set[IndexedSeq[Int]] = new Set[IndexedSeq[Int]]{
+    def contains(elem: IndexedSeq[Int]): Boolean = elem.zipWithIndex.forall{ idx => {
+        idx._2 >= 0 && idx._2 < shape(idx._1)
+      }
+    }
+
+    def +(elem: IndexedSeq[Int]): Set[IndexedSeq[Int]] = throw new UnsupportedOperationException(" member '+' is not supported in TensorN.keySet() ")
```
dlwh: this can just coerce to a set with `iterator.toSet + elem`

sujithjay:
Thank you for reviewing the PR, @dlwh . I hope to make the relevant changes, according to the review comments, sometime this week.

PR #700, opened 2018-03-15
Removed scalaxy.debug

superbobry:
Two reasons: it has been deprecated for a while (with Scalatest macros
suggested as a replacement), and it has been used in just a few places
in the Breeze codebase, and most of the call sites provide a useful
error message even w/o scalaxy.debug.


dlwh:
No, we need macro assertions for performance. We got a literal 100x speedup on small vector dot products from doing this. I'd be happy to replace them with other macro assertions, but it's not for pretty error messages.

superbobry:
We could release Breeze compiled with `-Xdisable-assertions` to make sure the assertions are excluded, wdyt? 

Also, why is the builtin `assert` slow? It is very likely to be inlined, and then it's a single test.

dlwh:
The thing that's slow is the by-name parameter in the RHS, which always allocates a closure (unless it gets stack allocated) I have real-world code where that stack allocation doesn't happen and it really did slow down actual code we were using by ~4-5x when small dot products were used in an inner loop.

I don't want to release without assertions. Those are precondition checks that I don't want to silently get ignored.

dlwh:
To be clear, this is only two argument assert/require/etc, but those are necessary to get good error messages (or, you know, macro assertions :-)

superbobry:
Oh... I'm actually surprised Scala compiler does not inline assertions automatically if they aren't disabled. 

In any case, what do you think about depending on Scalatest? :D

dlwh:
Sorry for being slow, as always.

I don't like that scalatest throws TestFailedExceptions, and it doesn't support require. I'm happy to consider another library, but I don't think it's the right fit

PR #721, opened 2018-08-13
Factor out DescriptiveStats.percentileInPlace

jklukas:
The percentile function always copies and sorts the input data, which provides good safety for the general case. This is inefficient, however, in cases where data already happens to be sorted or where the user wants to calculate a series of percentiles on a single dataset.

We factor out the sorting step and the percentile calculation step here and make a public percentileInPlace function that takes in a pre-sorted array so that users can opt in to the responsibility of array creation and sorting on their own.

Note that I'm relatively new to scala breeze, so very open to feeback on naming, documentation conventions, etc. If this simply doesn't fit in with the overall API you want to present to users, that's a perfectly fine outcome as well. Thanks for your work on this library!

### DIFF
--- a/math/src/main/scala/breeze/stats/DescriptiveStats.scala
+++ b/math/src/main/scala/breeze/stats/DescriptiveStats.scala
@@ -518,12 +518,24 @@ object bincount extends UFunc {
 object DescriptiveStats {
 
   /**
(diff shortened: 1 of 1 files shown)


dlwh:
thank you for doing this. sorry for being slow

jklukas:
Thanks, @dlwh!

PR #730, opened 2018-10-31
Add of a root finding method (as Matlab fzero)

abertout:
Implementation of a scalar root finding method similar to Matlab [fzero](https://fr.mathworks.com/help/matlab/ref/fzero.html). The method is based on the Brent's algorithm and comes along with other classical algorithms (Newton-Raphson, secant and bisection). Tests are provided.

### DIFF
--- a/math/src/main/scala/breeze/optimize/RootFinding.scala
+++ b/math/src/main/scala/breeze/optimize/RootFinding.scala
@@ -9,6 +9,7 @@ import scala.annotation.tailrec
 
 object RootFinding {
   lazy val eps = math.ulp(1d)
+  lazy val defaultMaxIter: Int = 10000
 
   /**
     * Generic method to compute a root approximation x of a function f such that f(x) = 0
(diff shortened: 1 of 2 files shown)


math/src/main/scala/breeze/optimize/RootFinding.scala
```diff
@@ -0,0 +1,147 @@
...
+    * @param fn function
+    * @param x0 first root estimate
+    * @param x1 optional second root estimate such that [x0,x1] brackets x
+    * @return
+    */
+  def find(fn: Double => Double, x0: Double, x1: Option[Double] = None): Double = {
+    //Generate a second estimation if needed
+    val xx1 = x1 match {
+      case Some(x) => x
+      case None =>  //search in an ever-widening range around the initial point (taken from octave fzero)
```
dlwh: i think this might be cleaner as something like `val xx1 = x1.orElse(findSecondEstimate(fn, x0).getOrElse(throw new RuntimeException(...))`. (I don't think IllegalArgumentException is right here. The user didn't provide bad arguments, the algorithm failed.)

abertout: I agree, I'm going to open a pull request with the correction.

Issue #731, opened 2018-11-09
bug in dot operator


BoleynSu:
val v = new DenseVector(Array(.1f,.2f,.3f,.4f,.5f,.6f,.7f))
val w = new DenseVector(Array(1.0f,1.0f,1.0f,1.0f,1.0f,1.0f,1.0f))
// v.dot(w) = 0

PR #732, opened 2018-11-09
fix issue #731

BoleynSu:
fix issue #731

### DIFF
--- a/math/src/main/java/breeze/linalg/operators/DenseVectorSupportMethods.java
+++ b/math/src/main/java/breeze/linalg/operators/DenseVectorSupportMethods.java
@@ -17,25 +17,26 @@ final public class DenseVectorSupportMethods {
      * @return
      */
     public static double smallDotProduct_Double(double[] a, double[] b, int length) {
-        double sum = 0.0;
+        double sumA = 0.0;
+        double sumB = 0.0;
         switch (length) {
             case 7:
-                sum = a[6] * b[6];
+                sumA = a[6] * b[6];
             case 6:
-                sum += a[5] * b[5];
+                sumB = a[5] * b[5];
             case 5:
-                sum += a[4] * b[4];
+                sumA += a[4] * b[4];
             case 4:
-                sum += a[3] * b[3];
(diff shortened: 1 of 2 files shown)


math/src/main/java/breeze/linalg/operators/DenseVectorSupportMethods.java
```diff
@@ -10,35 +10,32 @@
...
+     * WARNING: only returns the right answer for vectors of length < MAX_SMALL_DOT_PRODUCT_LENGTH
      * @param a
      * @param b
      * @param length
      * @return
      */
     public static double smallDotProduct_Double(double[] a, double[] b, int length) {
-        double sumA = 0.0;
-        double sumB = 0.0;
+        double sum = 0.0;
```
dlwh: please revert this. this code is deliberately written this way to be faster.

math/src/main/java/breeze/linalg/operators/DenseVectorSupportMethods.java
```diff
@@ -87,6 +84,8 @@ public static double dotProduct_Double(double[] a, int aoff, double[] b, int bof
     public static float smallDotProduct_Float(float[] a, float[] b, int length) {
         float sum = 0.0f;
         switch (length) {
+            case 7:
```
dlwh: thanks for catching this

dlwh:
could you add a test case please

math/src/main/java/breeze/linalg/operators/DenseVectorSupportMethods.java
BoleynSu: I removed sumA and sumB to make smallDotProduct for both double and float the same. If I revert this, should I use the same optimization for float?

dlwh:
ideally we'd benchmark it, but i think that's a good idea.

Thanks,
David

BoleynSu:
@dlwh I have added tests and reverted sum back to sumA and sumB.

dlwh:
Thanks!

PR #736, opened 2018-11-30
Fix dimension inconsistency in QR decomposition

alois-bissuel:
When doing a reduced QR decomposition of A (m x n matrix) in QR, where Q is m x k and R is k x n, and k = min(m,n), if m < n, then R is not square.
When asking for justR, the matrix produced by upperTriangular is square, hence it is incorrect


dlwh:
thanks!

PR #737, opened 2018-12-03
Remove unnecessary type bounds from Vector fold and scan functions.

### LINKED ISSUE #720: breeze DenseVectors (and possibly others) have bounded types in foldLeft.
I was looking at the [definition of `foldLeft` for `Vector`](https://github.com/scalanlp/breeze/blob/58b3f511f20521cdae0c09429ad278e8bec5a7f6/math/src/main/scala/breeze/linalg/Vector.scala#L114)

```
def foldLeft[B >: V](z: B)(op: (B, V) => B): B = valuesIterator.foldLeft(z)(op)
```

Comparing with the [signature for basic `foldLeft`, such as for `List`](https://www.scala-lang.org/old/node/12050.html), the basic implementation is fully generic over the accumulator output type.

In fact, this seems to be one of the specific reasons someone might be directed to use `foldLeft` over `fold`, when you want the accumulator output type to be arbitrary and potentially unrelated to the element type of the underlying container.

For e.g. `DenseVector` this can be a huge pain. Consider trying to count the number of elements where two different `DenseVector`s are equal.

```
scala> val x = DenseVector(1.0, 2.0, 3.0)
x: breeze.linalg.DenseVector[Double] = DenseVector(1.0, 2.0, 3.0)

scala> val y = DenseVector(3.0, 2.0, 1.0)
y: breeze.linalg.DenseVector[Double] = DenseVector(3.0, 2.0, 1.0)

scala> val w = (x :== y)
w: breeze.linalg.BitVector = BitVector(1)


scala> w.toDenseVector // note: fold over BitVector was even harder to deal with
res114: breeze.linalg.DenseVector[Boolean] = DenseVector(false, true, false)

scala> w.toDenseVector.map(i => if(i) 1 else 0).sum
res115: Int = 1

scala> w.toDenseVector.foldLeft(0)( (acc, i) => if(i) acc + 1 else acc)
<console>:15: error: type mismatch;
 found   : Int(1)
 required: String
       w.toDenseVector.foldLeft(0)( (acc, i) => if(i) acc + 1 else acc)
```

But folding over a `List[Boolean]` to get back a count of the `true` values, for example, is a perfectly standard operation.

Why does `Vector` have this type bound limitation in `foldLeft` (and presumably other operations, and perhaps for matrix types and other types)?

espears4sq:
This PR fixes #720 by removing unrequired type bounds in `foldLeft`, `foldRight`, `scanLeft` and `scanRight` for `Vector`.

The change compiles without error via `sbt compile` and `sbt test` passes all unit tests. Note: I did need to increase sbt memory to ~6GB or else I would occasionally have an OOM error during `sbt compile`. (This happens with just the unchanged repository though and I cannot see how it would be related to this set of changes).

Here is an example of the functionality working as expected:

```
scala> val v = DenseVector[Double](1, 2, 3, 4, 5)
v: breeze.linalg.DenseVector[Double] = DenseVector(1.0, 2.0, 3.0, 4.0, 5.0)

scala> v.foldLeft(false)((acc, v) => acc || (v > 4))
res5: Boolean = true
```

Here is an example of the same functionality failing if the change is not made:

```
scala> val v = DenseVector[Double](1, 2, 3, 4, 5)
v: DenseVector[Double] = DenseVector(1.0, 2.0, 3.0, 4.0, 5.0)

scala> v.foldLeft(false)((acc, v) => acc || (v > 4))
cmd4.sc:1: value || is not a member of AnyVal
val res4 = v.foldLeft(false)((acc, v) => acc || (v > 4))
                                             ^
Compilation Failed
```

### DIFF
--- a/math/src/main/scala/breeze/linalg/Vector.scala
+++ b/math/src/main/scala/breeze/linalg/Vector.scala
@@ -111,11 +111,11 @@ trait Vector[@spec(Int, Double, Float) V] extends VectorLike[V, Vector[V]] {
 
   /** See [[scala.collection.mutable.ArrayOps.foldLeft]].
    */
-  def foldLeft[B >: V](z: B)(op: (B, V) => B): B = valuesIterator.foldLeft(z)(op)
+  def foldLeft[B](z: B)(op: (B, V) => B): B = valuesIterator.foldLeft(z)(op)
 
   /** See [[scala.collection.mutable.ArrayOps.foldRight]].
    */
-  def foldRight[B >: V](z: B)(op: (V, B) => B): B = valuesIterator.foldRight(z)(op)
+  def foldRight[B](z: B)(op: (V, B) => B): B = valuesIterator.foldRight(z)(op)
 
   /** See [[scala.collection.mutable.ArrayOps.reduce]].
    */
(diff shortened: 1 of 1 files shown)


espears4sq:
Closing and re-opening to trigger a Travis rebuild.

espears4sq:
It seems something is breaking in the Travis build for Scala 2.10.6,

```
/home/travis/build/scalanlp/breeze/math/src/test/scala/breeze/linalg/maxTest.scala:31:15: diverging implicit expansion for type breeze.linalg.min.Impl[breeze.linalg.BroadcastedRows[breeze.linalg.DenseMatrix[Int],breeze.linalg.DenseVector[Int]],VR]
```

I am not able to reproduce it locally on my machine, but I am happy to work on a fix if this is related to the Vector type signature change in this PR.

dlwh:
thanks! I don't think we need tests or anything for this since it's simple.

don't worry about the travis build. that's on me.

Issue #706, opened 2018-04-10
Enhancement: avoid iterations over NaNs in StrongWolfeLineSearch


spockoyno:
According to Logger INFO, the loop in `def zoom(...)`  in `StrongWolfeLineSearch` keeps iterating 
`maxZoomIter` times, despite all the variables becoming NaN. Perhaps termination at this point would be useful?

Here's an example Logger output: 

> 16:54:49.618 [ScalaTest-run-running-TestAugmentedLagrangian] INFO breeze.optimize.StrongWolfeLineSearch - Line search t: 4.374953830356382E-13 fval: 222.99999996809137 rhs: 222.99999996984118 cdd: -3999.9930397465314
16:54:49.618 [ScalaTest-run-running-TestAugmentedLagrangian] DEBUG breeze.optimize.StrongWolfeLineSearch - hi=c
16:54:49.618 [ScalaTest-run-running-TestAugmentedLagrangian] INFO breeze.optimize.StrongWolfeLineSearch - Line search t: 4.37495383035638E-13 fval: 222.99999996809137 rhs: 222.99999996984118 cdd: -3999.9930397465314
16:54:49.618 [ScalaTest-run-running-TestAugmentedLagrangian] DEBUG breeze.optimize.StrongWolfeLineSearch - hi=c
16:54:49.618 [ScalaTest-run-running-TestAugmentedLagrangian] INFO breeze.optimize.StrongWolfeLineSearch - Line search t: 4.3749538303563794E-13 fval: 222.99999996809137 rhs: 222.99999996984118 cdd: -3999.9930397465314
16:54:49.618 [ScalaTest-run-running-TestAugmentedLagrangian] DEBUG breeze.optimize.StrongWolfeLineSearch - hi=c
16:54:49.618 [ScalaTest-run-running-TestAugmentedLagrangian] INFO breeze.optimize.StrongWolfeLineSearch - Line search t: NaN fval: NaN rhs: NaN cdd: NaN
16:54:49.618 [ScalaTest-run-running-TestAugmentedLagrangian] DEBUG breeze.optimize.StrongWolfeLineSearch - low=c
16:54:49.618 [ScalaTest-run-running-TestAugmentedLagrangian] INFO breeze.optimize.StrongWolfeLineSearch - Line search t: NaN fval: NaN rhs: NaN cdd: NaN
[...]

PR #747, opened 2019-05-21
throw exception as soon as t is NaN

### LINKED ISSUE #733: CubicLineSearch#interp produces `t = NaN`
This isn't easy to reproduce, but I'm running L-BFGS-B and at this step in CubicLineSearch#interp
```
val d2 = sqrt(d1 * d1 - l.dd * r.dd)
```
`d2` becomes NaN because `l.dd * r.dd` > `d1 * d1`. This only happens when fitting on certain datasets. Could you explain what this means in the context of the cubic interpolation? Any suggestions on how to safeguard against this?

danoliv:
Fix #733 and #706. Line search should not return NaN.


danoliv:
Any news on this MR ?

Issue #664, opened 2017-10-25
FunctionValuesConverged with relative = true


siqil:
For checking the convergence based on function values, we have the condition:

`info.length >= 2 && (state.adjustedValue - info.max).abs <= tolerance * (if (relative) state.initialAdjVal else 1.0)`

My understanding is that it checks whether the absolute difference between the recent maximum and current adjusted value is less than or equal to the tolerance scaled by the initial adjusted value (for relative = true). If that's correct, then imagine a function having negative values from the initial point till the minimum. Then this condition can never be satisfied, right? Shouldn't we use abs also on the RHS of the inequality?

dlwh:
yep, good catch!

PR #764, opened 2019-10-29
FirstOrderMinimizer#defaultConvergenceCheck fails with negative function values

Antti-Paladin:
I noticed that the default convergence check for `FirstOrderMinimizer` does not use absolute value of the function, when doing a check for relative gradient. I think when minimizing functions that yield negative values at the current considered point, the tolerance for gradient norm always falls back to `1E-8`.

Fixed to use the absolute value of the optimized function. Could you please check if this could be merged?

EDIT: Seems the CI pipeline doesn't pass, but the tests passed locally for me.


Antti-Paladin:
Has someone taken a look at this? I think it's very similar to https://github.com/scalanlp/breeze/issues/664, just this happens when checking the gradient.

PR #769, opened 2020-01-13
bump to Scala 2.13.1 (was 2.13.0)

SethTisue:


SethTisue:
the 2.11 job is failing — I guess that doesn't have to do with the changes in this PR?

the 2.13 job is failing with errors I'm also seeing in the Scala community build — any insight?

SethTisue:
ping, any maintainers around?

dlwh:
sorry, I'll try to get to this this week

SethTisue:
thanks!

I should probably clarify my intention here: anytime a project is unable to upgrade to a new Scala version, we are concerned that a regression in Scala might be at fault

since this isn't my codebase, I can't judge the actual likelihood of that myself. but you'll likely want or need to upgrade at some point anyway, so you'll probably need to dig into this eventually regardless.

dlwh:
I see. I think i broke something in my 2.13 build prior to your PR and
haven't gotten around to fixing it.

I was only building for 2.12 locally, and IIRC when I saw it was another
collection migration thing I just didn't want to deal with it at the time.
(It's not a bug in 2.13, just the API breakage for implementors of
collections and needing to have two versions of a class to support both
2.12 and 2.13)

dlwh:
thank you!

PR #786, opened 2020-06-20
fixes #785: use adjusted gradient and value in convergence check in FirstOrderMinimizer

### LINKED ISSUE #785: Adjusted values should be used in convergence update in FirstOrderMinimizer
As per https://groups.google.com/forum/#!topic/scala-breeze/YapqCs6222w `adjValue` and `adjGrad` should be used in convergence check in `FirstOrderMinimizer::infiniteIterations`

shvimas:
I've tried to come up with a suitable test but all my ideas seemed too convoluted for such a small change


dlwh:
thanks!

PR #794, opened 2020-10-31
Fix vector builder when building sparse vector with unique indices

darkjh:


darkjh:
Hello @dlwh can you take a look?

dlwh:
sorry about that!

PR #795, opened 2020-11-07
upgrade to ScalaCheck 1.15.1

SethTisue:
one tiny source adjustment needed, to adapt to typelevel/scalacheck#667

it would be convenient for me in the Scala community build if this were merged


SethTisue:
I believe (based on repeated local runs) that the CI failure is intermittent and unrelated to my change.

dlwh:
yeah sorry, it's mostly floating point noise that I've repeatedly tried to stomp out, but failed.

dlwh:
thanks!

PR #800, opened 2021-02-01
WIP -- added placeholder file for the nearPD function

ewalsh:
Adding a function to find the nearest positive definite matrix via the Higham (2002) algorithm.

### DIFF
--- a/math/src/main/scala/breeze/linalg/functions/nearpd.scala
+++ b/math/src/main/scala/breeze/linalg/functions/nearpd.scala
@@ -1,15 +1,5 @@
-// WIP file holder for a find nearest positive definite matrix function
+package breeze.linalg
 
-package breeze.linalg.functions
-
-import breeze.linalg.DenseMatrix
-import breeze.linalg.DenseVector
-import breeze.linalg.eigSym
-import breeze.linalg.fliplr
-import breeze.linalg.reverse
-import breeze.linalg.diag
-import breeze.linalg.*
-import breeze.linalg.Axis
 import util.control.Breaks.break
 
 /** Generate the nearest (in Frobenius norm) positive definite matrix
@@ -26,8 +16,10 @@ import util.control.Breaks.break
   * applied to the result of the Higham algorithm.
   * @param doSym Boolean indicating if X <- (X + t(X))/2 should be done,
   * after X <- tcrossprod(Qd, Q); some doubt if this is necessary.
-  * @param doDykstra Boolean indicating if Dykstra's correction should be used;
-  * true by default. If false, the algorithm is basically the direct fixpoint
+  * @param algo Algorithm sealed trait indicating which algorithm to use
+  * Option 1: Dykstra -- Indicating the use of Dykstra's correction;
+  * true by default.
+  * Option2: FixedPoint -- The algorithm is basically the direct fixpoint
   * iteration Y(k) = P_U(P_S(Y(k-1))).
   * @param onlyVals Boolean if TRUE, the result is just the vector of
   * eigenvalues of the approximating matrix.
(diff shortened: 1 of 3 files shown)


math/src/main/scala/breeze/linalg/functions/nearpd.scala
```diff
@@ -0,0 +1,276 @@
...
+  * using "F" is more in line with Higham's proposal.
+  * @param trace Boolean specifying if convergence monitoring should be traced.
+  */
+/**
+ * TODO
+ * migrate to breeze-y style, integrate with library, add tests
+ * @author ewalsh
+ **/
+
+class NearPD(matA: DenseMatrix[Double], corrBool: Boolean = false, keepDiag: Boolean = false,
```
dlwh, line 52: Is there a reason this needs to be a class? AFAICT you're only doing it to make it top level? Could you maybe put it in the linalg package object as a def? or make it an object with an apply method?

dlwh: (I don't think there's a reason to make this a "UFunc" fwiw)

math/src/main/scala/breeze/linalg/functions/nearpd.scala
```diff
@@ -0,0 +1,276 @@
...
+  * typically an approximation to a correlation or covariance matrix.
+  * If x is not symmetric (and ensureSymmetry is not false), symmpart(x) is used.
+  * @param corrBool Boolean indicating if the matrix should be a correlation matrix.
+  * @param keepDiag Boolean generalizing corr: if true, the resulting matrix
+  * should have the same diagonal (diag(x)) as the input matrix.
+  * @param do2eigen Boolean indicating if a posdefify() eigen step should be
+  * applied to the result of the Higham algorithm.
+  * @param doSym Boolean indicating if X <- (X + t(X))/2 should be done,
+  * after X <- tcrossprod(Qd, Q); some doubt if this is necessary.
+  * @param doDykstra Boolean indicating if Dykstra's correction should be used;
```
dlwh, line 29: i'd prefer there be an `algo: NearPD.Algorithm` or something, where NearPD.Algorithm is a sealed trait or an enum

math/src/main/scala/breeze/linalg/functions/nearpd.scala
```diff
@@ -0,0 +1,276 @@
...
+  do2eigen: Boolean = true, doSym: Boolean = false, doDykstra: Boolean = true,
+  onlyVals: Boolean = false, eigTol: Double = 1e-6, convTol: Double = 1e-7, posdTol: Double = 1e-8,
+  maxit: Int = 100, convNormType: String = "I", trace: Boolean = false) {
+  // ensure symmetry
+  val symA = (matA + matA.t) / 2.0
+  val diagX0 = diag(symA)
+  //--- start algorithm ---
+  def generate: DenseMatrix[Double] = {
+    // choose Dykstra or direct
+    val matX = chooseAlgo(symA, doDykstra)
```
dlwh, line 62: I don't like this. how about inlining?

`val matX = if(algo == Dystrka) doDystrkaFunc( ) else doDirect()`

math/src/main/scala/breeze/linalg/functions/nearpd.scala
```diff
@@ -0,0 +1,276 @@
...
+  // bool to double
+  def bool2Dbl(bool: Boolean): Double = {
+    if(bool){
+      1.0
+    } else {
+      0.0
+    }
+  }
+  // keepDiag function
+  def keepDiagFunc(mat: DenseMatrix[Double], bool: Boolean): DenseMatrix[Double] = {
```
dlwh, line 132: as with chooseAlgo i'd rather you just inline the boolean gating of this method and this method only does the keepDiag part

math/src/main/scala/breeze/linalg/functions/nearpd.scala
```diff
@@ -0,0 +1,276 @@
...
+              mat(i, j)
+            }
+          }
+        })
+      }
+     case false => mat
+    }
+  }
+  // doSym function
+  def doSymFunc(mat: DenseMatrix[Double], bool: Boolean): DenseMatrix[Double] = {
```
dlwh, line 149: same

math/src/main/scala/breeze/linalg/functions/nearpd.scala
```diff
@@ -0,0 +1,276 @@
...
+    bool match {
+      case true => (mat * mat.t) / 2.0
+      case false => mat
+    }
+  }
+  // corr diag function
+  def corrDiagFunc(mat: DenseMatrix[Double], bool: Boolean): DenseMatrix[Double] = {
```
dlwh, line 156: same

math/src/main/scala/breeze/linalg/functions/nearpd.scala
```diff
@@ -0,0 +1,276 @@
...
+    bool match {
+      case true => {
+        DenseMatrix.tabulate(mat.rows, mat.cols)({
```
dlwh, line 159: i might be misremembering, but I think this would be simpler/quicker as

```
val corrDiaged = mat.copy
diag(corrDiaged) := 1.0
corrDiaged
```

math/src/main/scala/breeze/linalg/functions/nearpd.scala
```diff
@@ -0,0 +1,276 @@
...
+  * @param onlyVals Boolean if TRUE, the result is just the vector of
+  * eigenvalues of the approximating matrix.
+  * the target vector to mean 0 and unit variance, default is false
+  * @param eigTol defines relative positiveness of eigenvalues compared to
+  * largest one, λ_1. Eigenvalues λ_k are treated as if zero when λ_k / λ_1 ≤ eig.tol.
+  * @param convTol convergence tolerance for Higham algorithm.
+  * @param posdTol tolerance for enforcing positive definiteness
+  * (in the final posdefify step when do2eigen is TRUE).
+  * @param maxit maximum number of iterations allowed.
+  * @param convNormType convergence norm type (norm(*, type)) used for Higham
```
dlwh, line 41: I'd rather a sealed trait here

math/src/main/scala/breeze/linalg/functions/nearpd.scala
```diff
@@ -0,0 +1,276 @@
...
+      })
+      matX = cpPt1 * matQtrim.t
+      dSubS = matX - matR
+      matX = doSymFunc(matX, doSym)
+      matX = corrDiagFunc(matX, corrBool)
+      matX = keepDiagFunc(matX, keepDiag)
+      conv = normTypeFunc(matX, matY, convNormType)
+      iter += 1
+      if(trace){
+        println(s"iter ${iter} : p = ${breeze.linalg.sum(vecd)} ||Y-X|| / ||Y|| = ${conv}")
```
dlwh, line 216: i prefer to use logging rather than printlns in breeze

math/src/main/scala/breeze/linalg/functions/nearpd.scala
```diff
@@ -0,0 +1,276 @@
...
+    var matY = matX
+    var matR = matY
+    while(iter < maxit && !converged){
+      matY = matX
+      matR = matY - dSubS
+      val eigVecVals = eigSym(matR)
+      val matQ = fliplr(eigVecVals.eigenvectors)
+      val vecd = reverse(eigVecVals.eigenvalues)
+      val p = vecd.map(x => x > eigTol)
+      val pcheck = breeze.linalg.sum(p.map(x => bool2Dbl(x)))
```
dlwh, line 198: val pcheck = any(vecd > eigTol)

math/src/main/scala/breeze/linalg/functions/nearpd.scala
```diff
@@ -0,0 +1,276 @@
...
+      if(pcheck < 1.0){
+        println("Matrix seems negative semi-definite")
+        break
```
dlwh, line 201: you have to wrap the loop you're breaking out of in breakable in Scala. It's dumb

math/src/main/scala/breeze/linalg/functions/nearpd.scala
```diff
@@ -0,0 +1,276 @@
...
+      }
+      converged = (conv < convTol)
+    }
+    if(!converged){
+      println(s"Did not converge in ${iter} iterations")
+    }
+    matX
+  }
+
+  def doDirectFixedPt(mat: DenseMatrix[Double]): DenseMatrix[Double] = {
```
dlwh, line 226: this function shares a ton of code with the one above. Is it possible to extra out more. I know I had you inline some stuff, but I think it would be cleaner to extract out the pcheck stuff, and the eigVecVals etc stuff, or possibly to make it so that the main loop takes a function as input that does the differences between the two algorithms, though I think that might be worse.

math/src/main/scala/breeze/linalg/functions/nearpd.scala
```diff
@@ -0,0 +1,276 @@
...
+    }
+    matX
+  }
+
+}
+
+// ## A longer example, extended from Jens' original,
+// ## showing the effects of some of the options:
+
+// val matA = DenseMatrix(Array(Array(1,     0.477, 0.644, 0.478, 0.651, 0.826),
```
dlwh, line 270: it would be good to make this into a test, and maybe add some randomized tests about properties this function should have, e.g.: for any matrix you give it (that one might want to give it), the output is a PD matrix or it errors, and it doesn't error too much, or something like that.

math/src/main/scala/breeze/linalg/functions/nearpd.scala
```diff
@@ -0,0 +1,276 @@
...
+  // choose algo
+  def chooseAlgo(mat: DenseMatrix[Double], bool: Boolean): DenseMatrix[Double] = {
+    bool match {
+      case true => doDykstraFunc(mat)
+      case false => doDirectFixedPt(mat)
+    }
+  }
+
+  // bool to double
+  def bool2Dbl(bool: Boolean): Double = {
```
dlwh, line 124: i don't think you actually need this here, but if you do, there's a function in breeze called `I` for indicator function that does exactly this, except it works on vectors

dlwh (changes requested):
sorry for being slow here, and thank you for submitting it. I have a bunch of comments about style, and I'd really like to see a few unit tests, since I really have no idea if this algorithm is correct.

math/src/main/scala/breeze/linalg/functions/nearpd.scala
ewalsh: First, thank you for your comments. I found all of them very helpful, and they are very much appreciated. It may take me a hot minute but I will address everything ASAP.

ewalsh: I have addressed or commented on each of these points. Please feel free to add any other comments, I will defer to you to revolve each based on your approval.

I have changed to an object with an apply method

math/src/main/scala/breeze/linalg/functions/nearpd.scala
ewalsh: I have added sealed traits for both the algorithm and choice of convergence norm type. This makes more sense and I hope is easier to read.

math/src/main/scala/breeze/linalg/functions/nearpd.scala
ewalsh: I have changed style to inline in each opportunity that I found

math/src/main/scala/breeze/linalg/functions/nearpd.scala
ewalsh: done

math/src/main/scala/breeze/linalg/functions/nearpd.scala
ewalsh: done

math/src/main/scala/breeze/linalg/functions/nearpd.scala
ewalsh: incorporated directly

math/src/main/scala/breeze/linalg/functions/nearpd.scala
ewalsh: done

math/src/main/scala/breeze/linalg/functions/nearpd.scala
ewalsh: This one is the only one I haven't truly addressed. The rationale is this, it would be possible to extract out a lot of the code here. However, when running the function it would require running an if statement within the loop to check which algorithm to run. 

Please advise. I'd be happy to change, but I just wanted to point out the original rationale. I think in large cases where lots of iterations are needed this repetitive checking could be slower than necessary.

math/src/main/scala/breeze/linalg/functions/nearpd.scala
ewalsh: Done and I also added another simple test

math/src/main/scala/breeze/linalg/functions/nearpd.scala
ewalsh: Agreed and removed

math/src/main/scala/breeze/linalg/functions/nearpd.scala
```diff
@@ -0,0 +1,252 @@
...
+                mat(i, j)
+              }
+            }
+          })
+        } else { mat }
+        out
+      }
+
+      // norm type function
+      def normTypeFunc(matX: DenseMatrix[Double], matY: DenseMatrix[Double], normType: ConvergenceNormType): Double = {
```
dlwh, line 153: i should really fix norm for matrices...

math/src/main/scala/breeze/linalg/functions/nearpd.scala
```diff
@@ -0,0 +1,252 @@
...
+          case Norm1 => breeze.linalg.max(breeze.linalg.sum(breeze.numerics.abs(matY - matX), Axis._0)) / breeze.linalg.max(breeze.linalg.sum(breeze.numerics.abs(matY), Axis._0))
+          case Norm2 => breeze.linalg.max(breeze.linalg.svd(matY - matX).singularValues) / breeze.linalg.max(breeze.linalg.svd(matY).singularValues)
+          case Infinity =>  breeze.linalg.max(breeze.linalg.sum(breeze.numerics.abs(matY - matX), Axis._1))  / breeze.linalg.max(breeze.linalg.sum(breeze.numerics.abs(matY), Axis._1)) // Infinity Norm
+        }
+      }
+
+      // break if negative semi-definite
+      def negSemiDefCheckBreak(pcheck: Boolean): Unit = {
+        if(!pcheck){
+          println("Matrix seems negative semi-definite")
```
dlwh, line 165: throw?

dlwh (commented):
sorry for being so slow, just a few more comments

PR #807, opened 2021-04-19
Make thread local field transient in `ThreadLocalRandomGenerator`

### LINKED ISSUE #806: ThreadLocalRandomGenerator not serializable
Below code:
```
val generator = new ThreadLocalRandomGenerator(new MersenneTwister())
val outputStream = new ByteArrayOutputStream(512)
val out = new ObjectOutputStream(outputStream)
try {
  out.writeObject(generator)
  outputStream.toByteArray
} catch {
  case ex: IOException => throw new SerializationException(ex)
} finally {
  if (out != null) out.close()
}
```
throws exception:
```
java.io.NotSerializableException: breeze.stats.distributions.ThreadLocalRandomGenerator$$anon$1
```

`ThreadLocalRandomGenerator` has value that is not serializabled (`ThreadLocal`)

rpiotrow:
* Make thread local field transient in `ThreadLocalRandomGenerator` to enable serialization
* Make it also lazy to not create ThreadLocal until it is used.

Fixes #806


dlwh (approved):
thanks!

Issue #715, opened 2018-06-20
Missing Transpose(DenseVector)*CSC operation?


vincent-grosbois:
Hello!

I have a CSC sparse matrix and a DenseVector.
I want to do the operation transpose(DenseVector)*CSCMatrix, which should yield a Transpose(DenseVector).
As far as I could see, this method is not implemented.
This seems implemented for transpose(SparseVector)*CSCMatrix (method `canMulSVt_CSC_eq_SVt` in CSCMatrixOps.scala ), but I could not find a Dense*CSC method.
Can someone confirm this implementation is not present in Breeze?
If so I add it

cheers

PR #809, opened 2021-04-29
Add method to multiply dense vectors with CSC matrices

aaronquantexa:
Hi, I saw this issue https://github.com/scalanlp/breeze/issues/715 and thought I'd have a go seeing as it's a good first issue (over 1 year old!).

I've copied the style of code around the repo and lightly cloned this test: "SparseVector * CSCMatrix Lifted OpMulMatrix & Transpose" to validate the matrix multiplication is correct for DenseVectors.

To allow the multiplication I've extracted 2 conversions (DenseVector to CSCMatrix and CSCMatrix to DenseVector - this latter one has an assertion inside that the Matrix has just 1 row), which are public.

Do let me know if there is more testing / other changes required.

### DIFF
--- a/math/src/main/codegen/breeze/linalg/operators/CSCMatrixOps.scala
+++ b/math/src/main/codegen/breeze/linalg/operators/CSCMatrixOps.scala
@@ -57,16 +57,17 @@ trait CSCMatrixOps extends CSCMatrixOps_Ring { this: CSCMatrix.type =>
     }
 
   implicit def canMulDVt_CSC_eq_DVt[T](
-      implicit op: OpMulMatrix.Impl2[CSCMatrix[T], CSCMatrix[T], CSCMatrix[T]],
+      implicit op: OpMulMatrix.Impl2[DenseMatrix[T], CSCMatrix[T], DenseMatrix[T]],
       zero: Zero[T],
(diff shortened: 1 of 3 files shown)


dlwh:
Hi @aaronquantexa, thanks for the PR!

Two things:

1) I think it makes more sense to get there via conversion from the DenseMatrix/CSCMatrix implementation from here https://github.com/scalanlp/breeze/blob/a1c56461b6b364eb4644a4ec1c5cdbc93b9bddec/math/src/main/codegen/breeze/linalg/operators/CSCMatrixOps.scala#L679 rather than the CSC/CSC, since a Transpose[DenseVector] is conceptually just a DenseMatrix with 1 row.

2) I'm in the middle of very large and obnoxious refactor to get Scala 3 support working, which has necessitated moving around a whole lot of stuff. If this doesn't merge cleanly into that branch, I may just copy and paste from your PR rather than merging, with due credit of course.

aaronquantexa:
Cool, I'll have a go at using DenseMatrix.

On point 2, sounds fair -  I'll leave this to you as maintainer.

aaronquantexa:
If it doesn't merge easily, can remove the additional - now redundant functions - in DenseVector.scala

jczestochowska:
related question: is subtraction between CSCMatrix and DenseVector supported? I tried `-` operator and all other related `-` operators but it seems like it's not implemented. @aaronquantexa @dlwh

PR #826, opened 2021-11-21
Eliminate cycle in DenseVector/HasOps initialization.

adampauls:
Addresses https://github.com/scalanlp/breeze/issues/825. I'm not sure if it fully eliminates any cycles, but it removes the observed deadlock in our code.


dlwh:
sorry for being slow! thanks adam!

PR #830, opened 2021-12-19
Updating Log4j version to 2.17.0 because of CVE-2021-45105 vulnerability.

jeremie1112:
Update Log4j version to 2.17.0
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-45105


dlwh:
Thanks! I'm not like super worried about it since it's a test dependency, but I appreciate it nonetheless!

jeremie1112:
No worries ! We always need to update to safety :smile:

PR #833, opened 2021-12-30
Wip  update spire

Quafadas:
This aims to remove the 2_13 dependancy on spire. Spire itself was compiled with 3.1.0 as far as I could tell, so I also bumped breezes scala version. 

That generated a bunch of problems. Upgrading the collection-compact modules, apparently fixed these problems.

With these changes, the tests in the "math" sub project passed (for me). 

Health Warnings:
Spire release is "M2". It may be preferable to wait for a 0.18.0
As I don't have a detailed understanding of Breeze, these changes may not be desirable in some larger context I'm not aware of.

### DIFF
--- a/math/build.sbt
+++ b/math/build.sbt
@@ -12,11 +12,11 @@ libraryDependencies ++= Seq(
   "org.apache.commons" % "commons-math3" % "3.2",
 //  ("com.chuusai" %% "shapeless" % "2.3.3").withDottyCompat(scalaVersion.value),
   "org.slf4j" % "slf4j-api" % "1.7.5",
-  ("org.typelevel" %% "spire" % "0.17.0").cross(CrossVersion.for3Use2_13),
+  "org.typelevel" %% "spire" % "0.18.0-M2",
   "org.apache.logging.log4j" % "log4j-slf4j-impl" % "2.17.0" % "test",
   "org.apache.logging.log4j" % "log4j-core" % "2.17.0" % "test",
   "org.apache.logging.log4j" % "log4j-api" % "2.17.0" % "test",
-//  ("org.scala-lang.modules" %% "scala-collection-compat" % "2.1.1").withDottyCompat(scalaVersion.value)
+  "org.scala-lang.modules" %% "scala-collection-compat" % "2.6.0"
 )
 
 // see https://github.com/typesafehub/scalalogging/issues/23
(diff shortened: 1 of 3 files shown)


dlwh:
Thanks! I have a slight preference for supporting 3.0.x, but if that's what Spire can support, than probably best to bump. I agree it's best to wait for a full release or at least RC before we cut a release with it.

armanbilge:
Just had a spire user run into the dreaded `Conflicting cross-version suffixes` trying to use spire and breeze together on Scala 3 and found my way here. Thanks for working on this!

> I have a slight preference for supporting 3.0.x

Thanks, this is helpful feedback. AFAIK nothing in spire depends on Scala 3.1.x (but can't be sure), so we can try rolling back if it's important to you. This issue has been controversial :) We could also wait and see about https://github.com/lampepfl/dotty/pull/14156, I think they originally promised it for 3.1.1 but I haven't followed closely.

> I agree it's best to wait for a full release or at least RC before we cut a release with it.

We released an M3 last week. I don't anticipate any more breaking changes going forward, we can probably RC or go straight to final.

romanowski:
lampepfl/dotty#14156 will be release in `3.1.2`. `RC1` should be release shortly and stable version 6 weeks later (unless there are no critical bugs).

Swoorup:
Likely also need to change:  https://github.com/scalanlp/breeze/blob/master/macros/build.sbt#L10

dlwh:
sorry for being slow. Yeah my preference for 3.0.x was driven by the lack of forwards compatibility, but if they're going to fix that, then that's great.

dlwh:
oops. shouldn't merge things when i have a migraine

dlwh:
(reverted the merge in master. don't know how to reopen a PR)

armanbilge:
> sorry for being slow. Yeah my preference for 3.0.x was driven by the lack of forwards compatibility, but if they're going to fix that, then that's great.

@dlwh do you still feel the same way about sticking to Scala 3.0.x? I think Spire will go straight to Scala 3.1. I experimented with the new [`-scala-output-version`](https://github.com/sbt/sbt/releases/tag/v1.7.0-M2) flag but it seems we've already used some Scala 3.1 APIs in macros so it doesn't work. I'd rather not backport these.

```
[error] -- Error: /workspace/spire/core/src/main/scala-3/spire/syntax/macros/literalMacros.scala:35:14 
[error] 35 |  parseNumber(digits.valueOrAbort.parts, BigInt(-128), BigInt(255)) match
[error]    |              ^
[error]    |method valueOrAbort was added in Scala release 3.1, therefore it cannot be used in the code targeting Scala 3.0
```

See some discussion on the topic in:
- https://github.com/typelevel/scalacheck/issues/878

dlwh:
i don't care that much anymore I guess, so fine with me!

PR #846, opened 2022-07-15
Bump dev.ludovic.netlib to 3.0.1

luhenry:
There are breaking changes in the package naming to support modules.


luhenry:
@dlwh :wave: I am in the process of upgrading `dev.ludovic.netlib` in Spark. But since 3.0.0 comes with a breaking change (change in the package names to accommodate modules), I need to get this change in Breeze as well.

Given that Spark is on org.scalanlp:breeze:2.0, I was also wondering whether it would be possible to do a 2.0.1 release with this patch as well? I'm happy to submit all the PRs necessary.

PR #857, opened 2023-12-30
single precision support added for LeastSquares calculation

dineshdharme:
Hello David Hall,

This PR adds support for doing Least Squares calculation for Float datatype.
Please review this.
```

  def NaiveMultiplicationFloat(n: Int, flag: Boolean): Unit = {


    val N_Dim = 4
    val rows = N_Dim + 5
    val columns = N_Dim
    val randomDoubleArray = Array.fill(rows * columns) {math.random}.map(ele => ele.toFloat)
    val sm: DenseMatrix[Float] = DenseMatrix.create(rows, columns, randomDoubleArray)

    val features2DMatrix = sm(::, 1 to -1)
    val targetPPNRVector = sm(::, 0)
   println("sm" +  sm)
    val result = leastSquares(features2DMatrix, targetPPNRVector)
    println("Breeze Linear Equation Solver results")


    println(s"intercept = ${result.coefficients.toArray.slice(1 , 5).mkString("[", ",", "]")}")
   println(s"r^2 = ${result.rSquared}")


  }
```

This code should allow you to test the Float datatype.


dlwh (changes requested):
Hi @dineshdharme thanks for submitting this!

So, typically in Breeze we aim for compile time errors rather than runtime errors. Here, we would get a runtime error if we tried to use it for Complex or something. so rather than the match, it would be better if there were two implementations of each of `matrixVector`, `matrixVectorSpecifiedWork` and `matrixVectorSpecifiedWork`

dineshdharme:
Thank you, I will update the PR tomorrow and let you know.

dineshdharme:
Hi David,

I made changes as you requested. Please take a look and let me know.

math/src/main/scala/breeze/stats/regression/LeastSquares.scala
```diff
@@ -37,69 +48,203 @@ private object leastSquaresImplementation {
...
-  def apply(X: DenseMatrix[Double]): DenseVector[Double] = X * coefficients
+
+
+
+case class LeastSquaresRegressionResult[T](coefficients: DenseVector[T], rSquared: T)(implicit ev: NumericType[T])
+  extends RegressionResult[DenseVector[T], T] {
+
+  def apply(x: DenseVector[T]): T = ev match {
+
+    case _: NumericType[Float]  =>
```
dlwh, line 97: let's make this method something like

```
def apply(x: DenseVector[T])(implicit can_dot: OpMulInner.Impl2[DenseVector[T], DenseVector[T], T]) = {
}
```

and then you can remove all of the switches and exception stuff.

math/src/main/scala/breeze/stats/regression/LeastSquares.scala
```diff
@@ -37,69 +48,203 @@ private object leastSquaresImplementation {
...
+      val x_ins = x.asInstanceOf[DenseVector[Float]]
+      (coeffs .dot(x_ins)).asInstanceOf[T]
+    case _: NumericType[Double]  =>
+      val coeffs = coefficients.asInstanceOf[DenseVector[Double]]
+      val x_ins = x.asInstanceOf[DenseVector[Double]]
+      (coeffs .dot(x_ins)).asInstanceOf[T]
+    case _ => throw new UnsupportedOperationException("Unsupported numeric type. Only Float and Double are supported")
+
+  }
+  def apply(X: DenseMatrix[T]): DenseVector[T] = ev match {
```
dlwh, line 108: similarly 

let's make this method something like

```
def apply(x: DenseVector[T])(implicit can_dot: OpMulMatrix.Impl2[DenseMatrix[T], DenseVector[T], T]) = {
}
```

and then you can remove all of the switches and exception stuff.

dlwh (commented):
almost there, thank you

dineshdharme:
I made the changes you requested. The only difference is for first apply I had to move the `implicit can_dot` to case class
as scala didn't recognize it to be `Function1`'s `apply` trait.  

Please review. thanks!

dlwh (approved):
thanks!

PR #859, opened 2024-07-01
Compile benchmark directory

bwignall:
The `benchmark` directory was not compiling on a fresh checkout. Some of this was because of changes to the Google Caliper dependency, some to Breeze APIs in the "root" directories, some other.

The `benchmark` directory now compiles cleanly, locally.

### DIFF
--- a/benchmark/src/main/scala/breeze/benchmark/BreezeBenchmark.scala
+++ b/benchmark/src/main/scala/breeze/benchmark/BreezeBenchmark.scala
@@ -18,12 +18,13 @@
 
 package breeze.benchmark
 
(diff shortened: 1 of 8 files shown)


dlwh (approved):
thanks, yeah I stopped running these on the regular. If you wanted to make some fast version of them run in CI that might be a good idea?

(As you can probably tell, Breeze is maintainence-mode with me flirting pressing the "archive this project" button. I haven't written non-trivial scala for 2.5 years.)

dlwh:
sure that's great. I'm gonna merge this as-is  if you want to follow up with more PR's, please lmk!

Issue #850, opened 2022-11-26
matrix multiply fails on transposed matrix, but succeeds on copy of transposed matrix


philwalk:
This seems like a bug to me, hopefully it's not a misunderstanding on my part.

The expression on line 38 compiles and executes as expected, but the seemingly identical expression on line 41 causes a runtime crash.  The main difference between the expression `X.t` and `(X.t).copy` AFAICT is the tranposed matrix has an offset of 1, due to the way it was created.

The behavior is the same with both scala 2.13.10 and scala 3.2.1.
I have tested in Windows 10 with jdk 11, and WSL Ubuntu with jdk 17 and scala 3.2.2-RC1, with the same results.

The code:

```scala
import breeze.linalg._

object ExpressionCrash {

  def main(args: Array[String]): Unit = {
    val yAndX = testmatrix
    printf("yAndX: %d x %d\n%s\n", yAndX.rows, yAndX.cols, yAndX)
    val y = yAndX(::,0) // column zero is y
    printf("y: %d\n%s\n", y.size, y)
    val X = yAndX(::,1 until yAndX.cols) // remainder of columns are predictors
    printf("X:  %d x %d\n%s\n", X.rows, X.cols, X)

    val N = X.cols.toDouble
    val Iᴛ: DenseMatrix[Double] = DenseMatrix.eye[Double](X.rows)
    val Iɴ: DenseMatrix[Double] = DenseMatrix.eye[Double](X.cols)

    val ιᴛ: DenseVector[Double] = DenseVector.ones[Double](X.rows)
    val ιɴ: DenseVector[Double] = DenseVector.ones[Double](X.cols)

    val T = X.rows.toDouble
    val Jᴛ: DenseMatrix[Double] = Iᴛ - 1.0/T * ιᴛ * ιᴛ.t
    val Jɴ: DenseMatrix[Double] = Iɴ - 1.0/T * ιɴ * ιɴ.t

    val ȳ = ιᴛ.t * y/T
    printf("ȳ: %s\n", ȳ)

    var Wxz: DenseVector[Double] = DenseVector.zeros(X.cols)
    val Z = DenseVector.ones[Double](X.rows)

    printf(" Jɴ dims: %s x %s\n", Jɴ.rows,  Jɴ.cols)
    printf("X.t dims: %s x %s\n", X.t.rows, X.t.cols)
    printf(" Jᴛ dims: %s x %s\n", Jᴛ.rows,  Jᴛ.cols)
[...]

PR #862, opened 2024-08-28
fix for #850: transpose-slice-multiply crash

philwalk:
The fix consisted of changing the comparison operator from `<` to `>` here:
diff math/src/main/codegen/breeze/linalg/operators/DenseMatrixOps.scala
```scala
       // if we have a weird stride...
       val a: DenseMatrix[Double] =
-        if (_a.majorStride < math.max(if (_a.isTranspose) _a.cols else _a.rows, 1)) _a.copy else _a
+        if (_a.majorStride > math.max(if (_a.isTranspose) _a.cols else _a.rows, 1)) _a.copy else _a
       val b: DenseMatrix[Double] =
-        if (_b.majorStride < math.max(if (_b.isTranspose) _b.cols else _b.rows, 1)) _b.copy else _b
+        if (_b.majorStride > math.max(if (_b.isTranspose) _b.cols else _b.rows, 1)) _b.copy else _b
```
The problem occurs when `majorStride` is too big rather than too small, so a `copy` is needed.
This might represent a performance improvement, on average, if it results in fewer calls to `_.copy`.

added a test to math/src/test/scala/breeze/linalg/DenseMatrixTest.scala
 ```scala
+  test("#850 - transpose slice multiply bug") {
+    val Jɴ = DenseMatrix( (0.75,  -0.25), (-0.25,  0.75))
+    val matB = DenseMatrix(
+      (67.0, 33.0),
+      (69.0, 78.0),
+      (93.0, 57.0)
+    ).t
+    val B = matB(::, 1 until matB.cols)
+    Jɴ * B      // should not crash
+  }
```

Also fixed a test that was failing in `Windows` due to a hardwired `/` path separator: `math/src/test/scala/breeze/linalg/TextOperationsTest.scala`
Fixed it by converting `\\` to '/' to find the correct relative path to  the resource file.

~~Upgraded the `sbt` version because `sbt` was crashing.~~

Upgraded scala versions.
```
diff --git a/project/Build.scala b/project/Build.scala
-  val buildCrossScalaVersions = Seq("3.1.3", "2.12.15", "2.13.8")
+  val buildCrossScalaVersions = Seq("3.4.2", "2.12.19", "2.13.14")
```

One test was failing on my system before I made any modifications, and is still failing now, so not sure what that's about.
```
[info] ProjectedQuasiNewtonTest:
[info] - optimize a simple multivariate gaussian (366 milliseconds)
[info] - optimize a simple multivariate gaussian with projection (136 milliseconds)
[info] - optimize a simple multivariate gaussian with l2 regularization (75 milliseconds)
[info] - optimize a complicated function without projection (393 milliseconds)
[info] - simple linear solve without projection (50 milliseconds)
L1 projection iter 67
[info] - simple linear solve with L1 projection compared to Octave *** FAILED *** (134 milliseconds)
[info]   1.1197254381828163E-4 was not less than 1.0E-4 (ProjectedQuasiNewtonTest.scala:221)
```

### DIFF
--- a/math/src/main/codegen/breeze/linalg/operators/DenseMatrixOps.scala
+++ b/math/src/main/codegen/breeze/linalg/operators/DenseMatrixOps.scala
@@ -301,9 +301,9 @@ trait DenseMatrixMultiplyOps extends DenseMatrixExpandedOps with DenseMatrixMult
 
       // if we have a weird stride...
       val a: DenseMatrix[Double] =
-        if (_a.majorStride < math.max(if (_a.isTranspose) _a.cols else _a.rows, 1)) _a.copy else _a
+        if (_a.majorStride > math.max(if (_a.isTranspose) _a.cols else _a.rows, 1)) _a.copy else _a
       val b: DenseMatrix[Double] =
-        if (_b.majorStride < math.max(if (_b.isTranspose) _b.cols else _b.rows, 1)) _b.copy else _b
+        if (_b.majorStride > math.max(if (_b.isTranspose) _b.cols else _b.rows, 1)) _b.copy else _b
 
       blas.dgemm(
         transposeString(a),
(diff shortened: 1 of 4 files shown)


dlwh:
thanks!

